PR #3, opened 2020-01-09
Update ressources.data.sncf.com.dmfr.json

elsa-pato:
Replace SNCF regional feed by SNCF national feed (previous feed was for trains in region "Pays de la loire" only, new feed is for all TER trains in France)


drewda:
@elsa-pato thanks for your contribution!

It looks good. However, GitHub Actions does not seem to be running the continuous integration checks. We're double checking with GitHub support on this. Will merge once that's fixed.

drewda:
Still no GH Actions, but we'll just go ahead anyway...

PR #20, opened 2020-02-04
WMATA

drewda:
from https://developer.wmata.com/


drewda:
Open question:

> We've been using the static feed at https://lrg.wmata.com/GTFS_data/google_transit.zip

> Is it necessary for us to switch to https://api.wmata.com/gtfs/bus-gtfs-static.zip and https://api.wmata.com/gtfs/rail-gtfs-static.zip ? Or is the lrg.wmata.com feed equivalent to the combination of the two hosted at api.wamta.com ?

PR #36, opened 2020-03-16
Update feed for BHTRANS (Brazil)

RockyTV:
I've updated the links for the GTFS feeds for BHTRANS (Empresa de Transportes e Trânsito de Belo Horizonte).

I don't know if it's a good idea to change the `id` and `feed_namespace_id` fields, but I've changed them to make it more readable as it looked like it was automatically generated. If there's a problem, I can easily revert back to what it was.

Also, the company provides GTFS files for the main lines, and additional lines, so in case I need to revert and readd those deleted fields, what would the `feed_namespace_id` value be for the second feed?


drewda:
@RockyTV thanks for your contribution!

Please change the feed ID back to its previous value. We use these Onestop IDs as stable, globally unique IDs.

You can keep the `feed_namespace_id` as-is and also add it to the new feed.

RockyTV:
Alright, I've reverted the changes to the feed ID. What about the ID for the other feed, should it follow the same name as the original one, changing the last character from an `a` to a `b` or keep it as is?

drewda:
`f-bhtrans~suplementar` is fine.

Two more notes: 

`  "license_spdx_identifier": "CDLA-Permissive-1.0"` is required for the DMFR format. If you want to specify the license for a feed, put that under each feed record. See https://github.com/transitland/distributed-mobility-feed-registry#basic-examples

You can also rename the file to `pbh.gov.br` now that the hostname has changed.

Thank you again.

RockyTV:
I've renamed the file to `dados.pbh.gov.br` to match the hostname and added the license to the feeds, as well as reverted back the schema license to `CDLA-Permissive-1.0`.

drewda:
@RockyTV perfect. Thank you for this contribution and for revising it so promptly.

RockyTV:
The feed was updated on the website but I don't see the additional feed I've added through the PR. Is there something wrong or is the site being refactored for the v2 atlas format?

PR #43, opened 2020-03-20
Operators

drewda:


adelcasse:
Associated feed for operator o-u-sncf shouldn't be f-u0-sncf~tram~train~ter but f-u0-sncf~ter following this update in Transitland Atlas https://github.com/transitland/transitland-atlas/pull/3

It might be the root cause of https://github.com/transitland/transitland-atlas/issues/58 ?

PR #61, opened 2020-06-16
Add GTFS-rt feeds for RTA Maryland

ericouyang:
Alternatively, see #62 for adding these feeds to a separate DMFR file for Swiftly

PR #62, opened 2020-06-16
Add Swiftly's GTFS-rt feeds for RTA Maryland

ericouyang:
Alternative PR to #61 using a separate DMFR.json file instead of the agency-specific one


drewda:
let's use #61 for now (unless in the future there are more feeds hosted on the swiftly domain)

ericouyang:
Thanks, @drewda. We do plan to add more of our customers' feeds into this over time. In that case, would you suggest aggregating them all into this one file (like trilliumtransit.com.dmfr.json), instead of agency-by-agency? If so, I'll incorporate your feedback from the other PR into this one and close that other one.

PR #63, opened 2020-06-16
Update TPER feed

eutampieri:


drewda:
@eutampieri thank you for updating these URLs. Do you know if there is a query parameter that will allow us to fetch the latest version every time, instead of hard-coding in a version number? If so, that would be ideal.

eutampieri:
I briefly checked and no, there isn’t

eutampieri:
Ok so the trick would be to fetch the URL from this page https://solweb.tper.it/web/tools/open-data/open-data-detail.aspx?source=&filename=gommagtfsbo

drewda:
Thanks for checking.

Good to know about the source of the URL for future reference (whether it's just manual checking or automatic scraping).

PR #74, opened 2020-07-11
Update static url for f-drce-kingstontransit

poldz123:


drewda:
@poldz123 thank you for this update

PR #68, opened 2020-06-18
Add Swiftly's GTFS-rt feeds for RTA Maryland

ericouyang:
Reopening #62 to create a DMFR.json file for Swiftly-provided realtime information

PR #75, opened 2020-07-15
Add Via Mobility GTFS-rt feeds

ericouyang:


drewda:
@ericouyang thanks for the contribution. Is there a URL you can add for `authorization.info_url`? That is, a place for users to sign up for auth keys.

ericouyang:
@drewda - Great suggestion! Just added that in.

Also, when you get a chance, could you help shed some light on the migration timeline in #68 ?

PR #79, opened 2020-07-25
Update static urls for bctransit

poldz123:


drewda:
@poldz123 thank you for these (and other) updates

poldz123:
@drewda I see that some of the feeds are not updated within the datastore. For example o-cbfg-winnipegtransit, its latest feed is not up-to-date but the static URL is.

drewda:
@poldz123 FYI, there is currently a once a day workflow that both pulls in new updates from Atlas and fetches feeds. I can't remember if this merged before or after the most recent workflow completed. If not, should sync fully shortly.

PR #80, opened 2020-07-25
Removed inactive wroute operator

poldz123:
Wroute already suspended the line: https://www.wroute.com/site/thankyou

PR #90, opened 2020-07-29
Replaced the amt and rtm and merge as one called exo

poldz123:
Amt and Rtm transit is one transit but for some reason, both were added which resulted in duplicated data.

Now they are merged as one and now called exo: https://exo.quebec/en/about/open-data


poldz123:
@drewda Not sure about the workflow error, can you guide me about it?

drewda:
@poldz123 looks good. thank you for all of these edits.

poldz123:
@drewda A follow up on this question: https://github.com/transitland/transitland-atlas/pull/80#issuecomment-665519753

Why is the Wroute is still in the datastore, should it be already removed from it?

poldz123:
@drewda Seems that the changes on this does not reflect in the feed registry in transit land: https://transit.land/feed-registry/?country=CA

PR #96, opened 2020-07-29
Update license info for VRS/KVB

mastacheata:
The VRS changed the license under which it publishes their public GTFS files and now dual licenses it as either ODC-By-1.0 or CC-By-4.0
I chose to set the license identifier to ODC because that's more easy to fit the database usage scheme of this use case and explicitly mentions exceptions granted by law for database-data.


drewda:
@mastacheata  Note that `license_spdx_identifier` describes the license attached to this DMFR file itself. Please instead add a property called `feed.license.spdx_identifier`: https://github.com/transitland/distributed-mobility-feed-registry#license Thank you.

PR #101, opened 2020-07-31
Update Metro Transit (MN) Feeds

botanize:
I renamed the feed file to reflect new feed location, updated feed locations, added realtime feeds, updated operator geohash (was outside of core Metro Transit service area).


drewda:
@botanize thank you for all of these additions (and good to see the public release of RT feeds).

Please keep the operator geohash as-is for the time being. (Transitland v1 had mechanisms for tracking the lineage of Onestop IDs as they changed over time, but v2 does not yet have all of this functionality. Until we add that addition, we're trying to keep Onestop IDs stable, even if the geohash component loses some meaning temporarily.)

botanize:
The validation doesn't provide enough information for me to diagnose why it's failing. Any ideas?

feeds/svc.metrotransit.org.dmfr.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "https://dmfr.transit.land/json-schema/dmfr.schema-v0.3.0.json",
+  "feeds": [
+    {
+      "spec": "gtfs",
+      "id": "f-9zv-twin~cities~minnesota",
+      "urls": {
+        "static_current": "https://svc.metrotransit.org/mtgtfs/gtfs.zip",
+        "static_planned": "http://svc.metrotransit.org/mtgtfs/next/gtfs.zip"
```
drewda, line 9: This is where the JSON Schema validation is currently failing:

```suggestion
        "static_planned": ["http://svc.metrotransit.org/mtgtfs/next/gtfs.zip"]
```

After this change, I think we should be ready to merge. Thanks.

botanize: Is there a place I can see the validation errors?

drewda: We don't yet run JSON Schema checks on GitHub Actions, but here's one way to run them locally:

```
wget https://dmfr.transit.land/json-schema/dmfr.schema-v0.3.0.json
npx ajv-cli test -s dmfr.schema-v0.3.0.json -d svc.metrotransit.org.dmfr.json --valid
```

drewda:
The most recent commit fixed the JSON Schema validation error. The file now fully validates at the schema level, but the feed validation script, which syncs the DMFR files into a local database for checking is failing in a way I have not seen before.

I just chatted with a colleague and it looks like there's actually a bug in the library code for loading from `feeds.languages[]`. Your PR is the first time the languages array has been populated in one of this repo's DMFR files.

We'll fix the library. But in the meantime, please just remove the two language arrays so we can merge this sooner rather than latter.

Thanks for helping us test some new parts of this process and for bearing with the delays!

PR #106, opened 2020-08-25
Operator cleanup

irees:


poldz123:
This MR causes most operators that are valid, example the canada operators all of them now does not have feeds anymore.

irees:
@poldz123 I am going to revert most of this PR today because of these issues.

PR #132, opened 2020-10-21
Added Arvin Transit and Alhambra Community Transit

rpedraza01:
Apologies for the many PRs lately. We've been implementing many new GTFS feeds for clients. There will probably be more to come later this month.


drewda:
@rpedraza01 not a problem. We can add feeds and PRs at any time.

PR #173, opened 2020-12-11
Create airpullmanspa.it.dmfr.json

airon90:
Created Airpullman SPA GTFS from http://www.agenziatpl.it/open-data/gtfs


drewda:
@airon90 thank you for this contribution.

Issue #172, opened 2020-12-11
Some GTFS data from Lombardy (Italy)


airon90:
Hi,
I have some GTFS data here: http://www.agenziatpl.it/open-data/gtfs
Only ATM is already available. I don't understand how to insert those data here so I leave this issue open hoping that someone will do this work :)

airon90:
I created #173 and #174, hoping that everything is correct. If so, I will finish the work later

drewda:
@airon90 thank you for all your contributions. Please consolidate into a single PR. Also please ensure that feed definition filenames end with `.dmfr.json`

airon90:
Sorry @drewda, I'm not good in git so I don't know how to rename and to merge all PR into one.

drewda:
At this point it's simplest to pick one of your existing PRs/branches and add to that.

Are you using github.com to make your edits or cloning the repo to your computer? Let me know which and I can provide some pointers.

airon90:
Github.com although I'd like someday to learn to use git itself. Keep in mind that I am not a real coder, just an enthusiast FOSS user with little coding skills.

On Github I just press the Edit button and do all the changes at once on a single file and then create a PR. Honestly I don't know if it is possible to make more edits at once on Github

drewda:
Thanks again for your contributions and apologies for the delayed response.

I've manually assembled all the feed additions into one PR at #208 I will close this issue now. But if I have forgotten to add any of the feeds, please feel free to reply and we can re-open the issue!

For future reference, it appears that you can start a PR the same way you did before. After you have done the first commit, go to the repository "fork" under your own user name, select the "branch" that is specific to that PR, and then use the "edit" button to further edit files. You can commit them in to the same branch and they will appear as additions to the PR. If you have questions about this in the future, feel free to let us know.

PR #179, opened 2020-12-13
Create f-movibus

airon90:
#172


airon90:
Be careful to rename the file, adding file extension :)

drewda:
Addressed by #208

PR #188, opened 2020-12-14
Added Union City Transit

rpedraza01:


drewda:
Transitland currently sources Union City Transit from 511.org — listing at https://www.transit.land/feeds/f-9q9jy-unioncitytransit I believe 511.org sources upstream from the Trillium feed, in which case we don't need this extra listing.

Will hold this PR temporarily while we double check.

rpedraza01:
@drewda you are right. Union City is sourced on TL from 511 which gets its GTFS feed from us at Trillium. I guess this PR can be removed or deleted, however that works.

drewda:
@rpedraza01 thanks for confirming. I'll close this PR.

PR #189, opened 2020-12-15
Add Interurban Trolley in Elkhart, IN

evansiroky:
Hat tip to https://www.transitwiki.org/TransitWiki/index.php/Publicly-accessible_public_transportation_data for a link to the MACOG feed.

PS... I couldn't find a link for South Bend Transpo...

### DIFF
--- a/feeds/macog.com.dmfr.json
+++ b/feeds/macog.com.dmfr.json
@@ -3,11 +3,11 @@
   "feeds": [
     {
       "spec": "gtfs",
-      "id": "f-bttgg-macog",
+      "id": "f-macog",
       "urls": {
         "static_current": " http://macog.com/google/google_transit.zip"
       },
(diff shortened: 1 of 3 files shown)


drewda:
@evansiroky thanks for this contribution.

Please edit the Onestop IDs for feed and operator. Your options:

1. replace `bttgg` with the geohash that refers to the bounds of the operator service area
2. use just a two-component Onestop ID like `f-macog`
3. if you meant `bttgg` to be part of the name component, rather than the optional geohash component, you can do `f-bttgg~macog` (tildes are allowed within the name component)

evansiroky:
I clearly didn't read the documentation enough. Requested changes submitted.

drewda:
Not a problem. We don't have detailed enough docs just yet, so it's helpful to see where people run into problems.

This looks good now.

PR #205, opened 2021-01-15
Add Régie intermunicipale de transport des Collines GTFS and GTFS-RT

brodyFlannigan:
This P/R adds the Transcollines (Ottawa area, Canada) Static and Realtime GTFS feeds.


drewda:
@brodyFlannigan thank you for this contribution

PR #213, opened 2021-01-25
Update and rename o-u0p-trentinotrasportieserciziospa.json

shoyip:
Updated the filename and the content of the json file due to change in company name and structure (see at https://www.trentinotrasporti.it/azienda/trentino-trasporti).


drewda:
@shoyip It is very useful to update name, short name, and URL. However, changing Onestop IDs may break some existing applications. Please revise this PR so that it only changes name, short name, and URL fields. Thank you.

drewda:
@shoyip thanks for the edits — and thanks for your patience — did not notice a notification about the changes.

PR #217, opened 2021-02-10
Removed Carson Circuit, added Telluride Galloping Goose

rpedraza01:


drewda:
@rpedraza01 do you know of a replacement for the Carson feed?

rpedraza01:
@drewda unfortunately, I don't know of a replacement for the Carson feed. From our communications with that agency, it appears very likely that Carson Circuit will be ceasing operations permanently by July. Currently, their service is completely suspended.

From the client:

> "...Our Council suspended Carson Circuit ops last March. Services are still suspended, and may not return. We are currently engaged in talks with Long Beach Transit, to have them operate brand new routes in Carson by July 1, 2021. Upon commencement, I presume their service planning team will take the lead on GTFS data. If not, I will touch base with you to discuss. Please let me know if I can be of assistance during the meantime."

drewda:
@rpedraza01 thanks for sharing those details.

Let me follow up on a way to tag the Carson feed for archiving, to keep the old feed versions on hand.

For the time being, I will remove that from this PR and just merge the addition of the new feed.

PR #224, opened 2021-02-28
Update dati.comune.lecce.it.dmfr.json

piersoft:
new endpoint on opendata municipality's platform


drewda:
@piersoft thanks, as always, for your contributions!

PR #230, opened 2021-03-26
Add feed_namespace_ids to Swiftly feeds

ericouyang:
### Problem
GTFS-realtime feeds are not showing up in Transitland V2 pages:
- https://www.transit.land/operators/o-9xj5s-viamobilityservices
- https://www.transit.land/operators/o-dqcq-regionaltransportationagencyofcentralmaryland

### Solution
I believe the issue is because `feed_namespace_id` is not set. `associated_feeds` is set, but appears insufficient.

This PR sets those values for the feeds

### DIFF
--- a/feeds/api.goswift.ly.dmfr.json
+++ b/feeds/api.goswift.ly.dmfr.json
@@ -4,6 +4,7 @@
     {
       "spec": "gtfs-rt",
       "id": "f-9xj5s-viamobilityservices~rt",
+      "feed_namespace_id": "o-9xj5s-viamobilityservices",
       "associated_feeds": [
         "f-9xj5s-viamobilityservices~co~us"
       ],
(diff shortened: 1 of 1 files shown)


ericouyang:
@drewda - Can you take a look at this PR? Thanks!

drewda:
@ericouyang thanks for looking into this and for your patience.

We've been looking into ways that static and real-time feeds can be associated. The current set-up is inconsistent and doesn't always work.

These additions should fix the relationships under the current set-up, so I'll merge it now.

We're sketching out a simpler replacement and will be rolling that out in the future to make these types of relationships easier to specify and maintain.

ericouyang:
thanks, @drewda ! I still don't see the feeds being associated together after this has merged. Can you take a look from your side at why this might be the case?

drewda:
@ericouyang yes, we're investigating, and we're working toward a more systematic solution under #248

PR #232, opened 2021-04-02
Added CityBus

rpedraza01:
I know CityBus already exists in the repo here https://github.com/transitland/transitland-atlas/edit/master/feeds/gocitybus.com.dmfr.json, but the client has confirmed that they want to use Trillium's GTFS feed URL going forward. Is this PR the best course of action to switch GTFS feed URLs?


drewda:
@rpedraza01 yes, this is a good overall approach for switching URLs. Here are two edits to make:

- use the same existing feed Onestop ID in the new record (`f-dp4j-citybus`); do not create a new feed Onestop ID
- delete the `gocitybus.com.dmfr.json` file in this PR, so when it is merged only one record for the feed will remain

If you need help with either, feel free to let me know.

rpedraza01:
@drewda I could use some help with the feed URL switch. I tried to add the original Onestop ID directly to `patch-43` but it failed. I'm now hesitant to do anything further or delete `gocitybus.com.dmfr.json`.

drewda:
@rpedraza01 the failure is fine for now. (It's just complaining that `f-dp4j-citybus` is currently duplicated in more than one place.)

To delete the `gocitybus.com.dmfr.json` file you can go to https://github.com/rpedraza01/transitland-atlas/blob/patch-43/feeds/gocitybus.com.dmfr.json and click the trash can icon in the upper right.

rpedraza01:
@drewda thanks! I deleted the `gocitybus.com.dmfr.json` file and it appears that this PR has now passed its checks and been accepted. I guess I should've deleted the file then make the Onestop ID change.

drewda:
@rpedraza01 looks good.

FYI, it's alright to have an incomplete PR or one that fails the checks temporarily while you are working on it, as long as it's fixed before we actually merge it.

PR #235, opened 2021-04-08
Added Porterville Transit

rpedraza01:


drewda:
@rpedraza01 looks like this should replace `f-9q7m-portervilletransit` currently defined in `tularecog.org.dmfr.json` -- can you follow the process again to remove that record and add that Onestop ID to your new addition with the Trillium URL?

rpedraza01:
@drewda I'm not sure why the checks failed this time. I added the original Onestop ID to Trillium's new URL and removed the record from `tularecog.org.dmfr.json`. Could you please help me again?

drewda:
@rpedraza01 your changes are both right, but the issue is that they need to be in the same branch and the same PR. Here are my suggestions for how to proceed:

1. open https://github.com/rpedraza01/transitland-atlas/tree/patch-46 (the branch associated with this PR)
2. navigate to the `tularecog.org.dmfr.json` file, remove the relevant record, and "commit" it to the existing branch
3. the tests should then pass on this PR

Separately, you can close https://github.com/transitland/transitland-atlas/pull/240 because that won't be needed.

Let me know if you have any more questions as you try this.

rpedraza01:
@drewda I think I followed your suggestions correctly but it still failed the checks. I removed the Porterville entry from `tularecog.org.dmfr.json` in `patch-46` but it still doesn't seem to like it.

I then tried merging `patch-46` into the most recent PR, `patch-49` then merging those two into the `master` branch (see #242 ). I'm not sure what to do.

feeds/trilliumtransit.com.dmfr.json
```diff
@@ -2522,6 +2522,12 @@
     },
     {
       "spec": "gtfs",
+      "id": "f-9q7m-portervilletransit",
+      "urls": {
+        "static_current": "http://data.trilliumtransit.com/gtfs/porterville-ca-us/porterville-ca-us.zip"
+      }
+    },
+    {
```
drewda, line 2530: @rpedraza01 try adding this missing part:
```suggestion
    {
      "spec": "gtfs",
```

drewda:
@rpedraza01 I located the specific issue above. I'll let you press "commit suggestion" yourself so you can see the results. (It will hopefully turn this to a passing build!)

FWIW, doing more than one edit using the GitHub web interface is a challenge for everyone. If you are up for trying an alternative, you may want to try downloading the GitHub Desktop client: https://desktop.github.com/ That way you can edit the files in your own text editor, delete them, and then decide when to make a commit and open a PR.

rpedraza01:
@drewda thank you so much for your help and patience. Your suggested fix appears to have fixed it. Could it have really been a line that was holding everything up? I would read through each line of changes looking for anything out of place and didn't notice that a `"spec": "gtfs",` was missing.

I'll think about downloading the repo to my machine to make edits and pushes. But it seems like the only time we encounter issues is when an agency already has their own `.json` file and an existing Onestop ID. What would be the best way to search for an existing agency in the repo?

drewda:
@rpedraza01 you're welcome!

Yes, working with the repo on your own machine is probably only necessary when you are going to edit more than one file for a PR.

Some options for searching for existing feeds:

- Search within the Atlas repository by using the search bar in the upper left of this screen
- Search on the Transitland website. Here is an overview of how to do that: https://www.transit.land/news/2021/03/04/transitland-v2-search You can then follow links back to the files in the Atlas repository, or you can copy out Onestop IDs and use them to search your local copy of the Atlas repo.

PR #271, opened 2021-07-12
Added Garden of the Gods

rpedraza01:


drewda:
@rpedraza01 thanks for fixing that!

PR #295, opened 2021-08-16
Updated fetch link for Harrisonburg VA

julieteldred:


drewda:
FYI, @julieteldred it looks like the test was failing because of an extra space at the beginning of the URL field. I removed the space and the tests now pass.

drewda:
Ah, I see this is what you did in #295

PR #302, opened 2021-08-19
Update URL and license

FDMS:
See https://www.data.gv.at/katalog/dataset/ab4a73b6-1c2d-42e1-b4d9-049e04889cf0


drewda:
Thanks for this contribution!

Please leave `license_spdx_identifier=CDLA-Permissive-1.0` — that applies to the DMFR file itself. You can set `feeds[] .license.spdx_identifier=CC-BY-4.0` For an example, see https://github.com/transitland/distributed-mobility-feed-registry#basic-examples

PR #336, opened 2021-09-20
Update URL for f-dqc-thebusofprincegeorgescounty

ethanpooley:
See the link on this page: https://www.princegeorgescountymd.gov/1122/Routes-Schedules

Unfortunately this URL will eventually change again, and also there are some inadequacies in the stop_times, at least as far as OpenTripPlanner is concerned.


drewda:
@ethanpooley thanks for the contribution!

PR #352, opened 2021-10-04
Trillium Feed Maintenance: Moved, deleted, fixed, and added feeds

KonaTrillium:
This PR has a variety of maintenance items on Trillium feeds. 

- Moved feeds hosted at oregon-gtfs.com to the correct DMFR file. 
- Deleted the former feeds for the Oregon POINT and added the new, singular feed containing these data. 
- Fixed an error in a feed URL. 
- Added several feeds we've been maintaining that were missing. 

I'm anticipating needing to change the new POINT feed—does that need to reference to the old feed IDs? These are now completely gone and will not be returning or otherwise valid in the future.


scripts/debug/add_feed_tags.py
```diff
@@ -63,26 +63,22 @@
     {"onestop_id":"f-c283-jamestownsklallam","tags":{"managed_by": "mjc"}},
     {"onestop_id":"f-c2qf-mountainline","tags":{"gtfs_data_exchange": "mountain-line"}},
     {"onestop_id":"f-9rb-caravan~or~us","tags":{"feed_id": "caravan-or-us", "managed_by": "trillium", "gtfs_data_exchange": "caravan-airport-transportation"}},
-    {"onestop_id":"f-9rb-cascadespoint~or~us","tags":{"feed_id": "cascadespoint-or-us", "managed_by": "trillium", "gtfs_data_exchange": "cascades-point"}},
```
drewda, line 66: No need to edit this file. (It's just left over from past data maintenance.)

feeds/oregon-gtfs.com.dmfr.json
```diff
@@ -1413,6 +1266,84 @@
           ]
         }
       ]
+    },
+    {
+      "spec": "gtfs",
+      "id": "f-point~or-us",
```
drewda, line 1272: We have been adding a new way to handle situations where one feed takes the place of multiple previous ones. You can add a `supersedes_ids` property with an array of the previous feed Onestop IDs. For example:
 https://github.com/transitland/transitland-atlas/blob/ca9e9eb40090b5da6a2b195f1990c5909027d6c4/feeds/avta.com.dmfr.json#L7

feeds/oregon-gtfs.com.dmfr.json
```diff
@@ -1413,6 +1266,84 @@
...
+      "urls": {
+        "static_current": "http://oregon-gtfs.trilliumtransit.com/gtfs_data/point-or-us/point-or-us.zip"
+      }
```
drewda, line 1275: Please also add an `operators` property with a single record for the new POINT operator. The operator record can also have a `supersedes_ids` property with an array of the old operator Onestop IDs.

If you have any questions about this, feel free to let us know — I know it's not yet well documented.

drewda (commented):
Thanks for all this maintenance. See comments and let us know if you have any questions.

scripts/debug/add_feed_tags.py
KonaTrillium: I thought it might be something like that, but took them out just to keep things consistent. 

New commit undoes this change.

KonaTrillium:
New commit undoes the removal of old IDs from the past data maintenance script and adds in the operator/superseded ID info.

drewda:
Looks good!

PR #353, opened 2021-10-09
Update Toronto TTC GTFS static feed.

mmathieum:


drewda:
Thanks for this update.

PR #354, opened 2021-10-10
Added Transports en Commun Lyonnais

gfanton:
Added Transports en Commun Lyonnais (TCL)
- https://www.tcl.fr
- https://transport.data.gouv.fr/resources/8378

### DIFF
--- a/feeds/download.data.grandlyon.com.dmfr.json
+++ b/feeds/download.data.grandlyon.com.dmfr.json
@@ -19,7 +19,7 @@
           "associated_feeds": [
             {
               "gtfs_agency_id": "1"
-            },
+            }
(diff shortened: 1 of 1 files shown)


drewda:
@gfanton thanks for this contribution. If you have time to add more feeds from transport.data.gouv.fr, we would appreciate the assistance.

gfanton:
@drewda maybe you can create a small script that can parse:
- GPRS: https://transport.data.gouv.fr/api/datasets
- GRPS-RT: https://tr.transport.data.gouv.fr/
and convert it to `dmfr.json` ? 
I think that If you can do that, you may have the whole country covered

PR #357, opened 2021-10-11
Create dados.natal.br.dmfr.json

VitorRDias:
Public feed of Natal, Rio Grande do Norte, Brazil


VitorRDias:
> 
> 
> @VitorRDias thank you for this submission.
> 
> `urls.static_current` cannot point to a directory of files; it must point to a ZIP archive. Is there a URL for a ZIP file for this feed?

I could not find a ZIP file for it; the open data website contains individual links to each file.

drewda:
@VitorRDias thanks for checking again.

For now, the [transitland-lib](https://github.com/interline-io/transitland-lib) tooling can only reference URLs that are ZIP files. In the future, we may be able to change this.

PR #376, opened 2021-11-11
Add GTFS feed for Icelandic bus company Strætó

flother:
Source data and licence doc available from <https://www.straeto.is/is/um-straeto/leidakerfi/opin-rafraen-gogn>.

Strætó is the publicly-owned company that operates buses within Reykjavík and its surrounding towns, along with a few other routes around the country.


drewda:
@flother thank you for this contribution.

PR #382, opened 2021-11-17
use 2021 data for swiss gtfs feed

izolight:


drewda:
Thank you for this contribution.

PR #414, opened 2022-01-14
Link to real (and updated) GTFS for RIOC

alexwb72:
I run the GTFS feed for RIOC - we store the feed at the National RTAP link above. This is the link referenced by Google et. al.

The one currently in there appears to be wrong and has not been updated in a number of years.


drewda:
We have been looking for an official feed for RIOC. Thanks for sharing this.

PR #415, opened 2022-01-19
Adding greensboro GTA feed record

brockhaywood:
Adding the GTFS schedule for https://www.greensboro-nc.gov/departments/transportation/greensboro-transit-agency-public-transportation-division


drewda:
@brockhaywood thank you for this contribution. We should combine together the feed records for `f-greensboro~nc~us` before merging this. Will do so in #416

PR #418, opened 2022-01-24
Update mjcaction.com.dmfr.json

evantrillium:
removed Curry Public Transit, moved to oregon-gtfs


evantrillium:
scrapping this - back to the drawing board!

PR #429, opened 2022-02-09
Add MVgo GTFS Realtime URLs

brendannee:


drewda:
@brendannee thank you for this contribution!

PR #436, opened 2022-02-15
standardize all entries

evantrillium:
updated urls
updated current/historical urls
added missing operators (v1 - more to come)
alpha organized

### DIFF
--- a/feeds/oregon-gtfs.com.dmfr.json
+++ b/feeds/oregon-gtfs.com.dmfr.json
@@ -1,1389 +1,1386 @@
 {
-  "$schema": "https://dmfr.transit.land/json-schema/dmfr.schema-v0.4.0.json",
-  "feeds": [
-    {
-      "spec": "gtfs",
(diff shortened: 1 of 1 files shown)


evantrillium:
@drewda is there a way I can see what issues (eg, missing " or incorrect syntax on an entry) exist? I am trying to get this as standardized as possible to reduce fails but except for ensuring all } and ] are closed I am not certain how to best check against the defined schema

drewda:
Hi @evantrillium, my colleagues and I use Visual Studio Code, which can automatically run the JSON file through the schema and report errors. Here's a screenshot showing the "problems" tab reporting errors and warnings in the oregon-gtfs.com.dmfr.json` file:

image

VS Code is available for download, or you can run it in a web browser at https://vscode.dev/

Or if you'd rather stick with your current editor, you can also run JSON schema validation using command-line tooling, like so: https://github.com/transitland/transitland-atlas/blob/master/.github/workflows/validate.yml#L25-L26

PR #443, opened 2022-02-17
Add Marin Transit GTFS Realtime URLs

brendannee:
Update Marin Transit static GTFS URL


drewda:
@brendannee thanks for adding this!

PR #447, opened 2022-02-28
Update Atlas to use GTFS feed hosted at Amtrak.com

e-lo:
As of Feb 27, 2022 Amtrak hosts the GTFS feed on their website.


drewda:
@e-lo thanks for this contribution (and also to all the Cal-ITP folks involved behind the scenes in helping to make this happen)

PR #450, opened 2022-03-09
Create autna.es.dmfr.json

simportela:
Public GTFS static data for AUTNA SL.
Thank you!


drewda:
@simportela thank you for this contribution. Is this an official feed? If so, is it possible to source it from autna.com instead of a Google Drive link?

simportela:
Hello Drew, this is the official feed source and it's the same used for Google Maps. We cannot source it from autna.com. Would this be a problem?

drewda:
@simportela thanks for confirming. Yes, we can work with this for now. If the URL changes in the future, please update it here or email us at hello@transit.land

PR #460, opened 2022-03-28
GTFS feeds of Siauliai and Alytus (Lithuania)

vycius:
Adding two more GTFS feeds of two Lithuanian cities:

- Siauliai
- Alytus


drewda:
@vycius thanks for this contribution.

We are still looking into why the validation process is failing. Perhaps there are some UTF8 characters that the validator does not like. Once we are able to figure out and fix the issue, we will merge.

vycius:
The problem has been fixed.

PR #495, opened 2022-05-10
Added Zagrebački električni tramvaj GTFS RT

NadaTheOptimist:
First time I've ever done anything like this, hope this is ok.


drewda:
@NadaTheOptimist thank you for this contribution. I have made a few small fixes to group the static and real-time feed under the single operator record.

Issue #211, opened 2021-01-20
GitHub Actions cronjob should open PRs, instead of just pushing directly


drewda:
A longer term fix for #171

PR #502, opened 2022-05-23
Lithuania's GTFS Realtime feeds

vycius:
- Added three GTFS Realtime feeds for Lithuanian municipalities;
- Split feeds and operators;
- Add new GTFS of Taurage region;


drewda:
@vycius thank you for these contribution.

I am not sure why the automated job tried to modify the GBFS and MDS files in this branch. We will have to look into this further as part of https://github.com/transitland/transitland-atlas/issues/211

PR #515, opened 2022-06-09
added City of Miami Beach, updated two now deprecated entities

evantrillium:


drewda:
Thanks, @evantrillium !

PR #546, opened 2022-07-01
Add GTFS-RT for Queensland, Australia

icjuego:
https://gtfsrt.api.translink.com.au/

### DIFF
--- a/feeds/translink.com.au.dmfr.json
+++ b/feeds/translink.com.au.dmfr.json
@@ -20,7 +20,24 @@
       },
       "license": {
         "spdx_identifier": "CC-BY-4.0",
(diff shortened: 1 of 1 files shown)


drewda:
@icjuego thank you for this contribution.

Queensland bus services have been reorganized and rebranded since we last updated this data, so I've tried to update the "operator" records to use the new brand names and groupings.

If you are familiar with Queensland and see any errors in the names or groupings, please let us know.

icjuego:
Thanks for fixing my errors/omissions and merging :)

Yes, I am familiar with most, they're correct now. At least until the next rebranding/reorg!

PR #548, opened 2022-07-01
Add PIGMA, with aggregated Nouvelle Aquitaine french regional GTFS feed

kalon33:
All infos are taken from GTFS' agency.txt file.

### DIFF
--- a/feeds/pigma.fr.dmfr.json
+++ b/feeds/pigma.fr.dmfr.json
@@ -262,15 +262,6 @@
           ]
         },
         {
-          "onestop_id": "o-ca_pays_basque-hegobus,
-          "name": "Hegobus",
-          "website": "https://hegobus.fr",
-          "associated_feeds": [
-            {
-              "gtfs_agency_id": "CA_PAYS_BASQUE:Operator:Hegobus"
-            }
-          ]
-        },        {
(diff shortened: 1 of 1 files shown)


kalon33:
Dear @drewda, would it be possible to have a look at this? In order to know if it's OK or if I should make some changes?

Thanks a lot for your help.

drewda:
@kalon33 this looks good. Two notes:

We use an "opinionated" format for the DMFR files in this repository. It's not necessary for you to do this before opening a PR. But for your information, the way I applied this to your file is the following command: `transitland dmfr format --save feeds/pigma.fr.dmfr.json`

It's nice but not necessary to create operator records for every agency in a feed. The Transitland platform can automatically generate virtual operator records. It's only necessary to create operator records when you want to link together static and realtime feeds, or when you want to provide additional attributes (such as tags). 

Thank you for all your contributions for French feeds.

kalon33:
@drewda thanks for your explanations. This GTFS feed covers a huge part of South West of France, and "embeds" notably data in data.bordeaux-metropole.fr.dmfr.json, but also others, that are currently marked as not updated for a long while (900, 1000 days or more). I imagine they should be removed? Or this is dealt "automatically" on your side?

PR #553, opened 2022-07-08
Update googletf.gohrt.com.dmfr.json

wtimmerman-fitp:
Updates static_current URL and moves previous static_current to static_historic

Current feed on transit.land has effective date through 2021-05-15. An updated feed URL found on https://virginia-gtfs.com/ (as I understand, part of a Virginia DRPT effort to centralize feed links for Virginia) has the agency's current feed, which goes through September 2022.

Not sure if the name of the json file needs to change given the source URL change. Edit instructions didn't seem to speak to that so I left as is.

### DIFF
--- a/feeds/gohrt.com.dmfr.json
+++ b/feeds/gohrt.com.dmfr.json
@@ -6,7 +6,9 @@
       "spec": "gtfs",
       "urls": {
         "static_current": "https://gtfs.gohrt.com/gtfs/google_transit.zip",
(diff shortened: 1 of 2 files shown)


drewda:
@wtimmerman-fitp thanks for this contribution!

The one change that was needed is that `feeds.urls.static_historic` takes an array, rather than a single URL.

The name of the DMFR is just a stylistic convention -- it's not required to rename it.

wtimmerman-fitp:
Got it, thanks for the fix!

PR #571, opened 2022-07-27
Update Tisséo to the new v2 dataset

fmauNeko:
Hello,

Tisséo released a new v2 set and deprecated the v1 set.
I updated the URL with the new dataset.

Have a good day !


drewda:
@fmauNeko thank you for this update

PR #572, opened 2022-07-27
Add Réseau Mistral

fmauNeko:
Hello,

Here's a new network, Réseau Mistral, from the city of Toulon in France.

Have a good day !


drewda:
@fmauNeko thank you for this contribution

PR #574, opened 2022-07-29
Updated LPP to use the new GTFS URL

franga2000:
Looks like the feed URL for *Ljubljanski Potniški Promet* (Ljubljana, Slovenia) has been down for a few years now. 

I've updated it to the correct URL.


drewda:
@franga2000 thanks for this update

PR #576, opened 2022-08-07
update Belgrade data to latest feed

stalker314314:


stalker314314:
If it works, you might use https://data.gov.rs/sr/datasets/r/729be9a1-7ed9-453d-9a3d-68fa30f07529 as an URL. It always points to latest feed (albeit you need to query it to see if it changed). If it works, it might be better to use that!

BTW, source of this data is here: https://data.gov.rs/sr/datasets/gtfs/ (Serbian only)

drewda:
@stalker314314 thank you for this contribution!

PR #581, opened 2022-09-01
Thailand, mostly Bangkok for now

oranv:
https://namtang.otp.go.th/opendata


drewda:
@oranv thanks for this contribution!

PR #585, opened 2022-09-21
update Broome County url from: https://511ny.org/developers/resources

szjozsef:


drewda:
@szjozsef thanks for contributing this update

PR #586, opened 2022-09-22
update url from: https://www.parkcity.org/departments/transit-bus/gtfs

szjozsef:


drewda:
@szjozsef thank you for this update

PR #593, opened 2022-10-04
Create rejseplanen.dk.dmfr.json

akaise21:
I used the Samtrafiken file as a template, as I couldn't find the example file. For now, I've just added the data provider as the operator. Should I add all operators/agencies instead? They are included in the GTFS file, so using those instead would make it easier to create and maintain them. Thanks!

### DIFF
--- a/feeds/rejseplanen.dk.dmfr.json
+++ b/feeds/rejseplanen.dk.dmfr.json
@@ -7,18 +7,6 @@
       "urls": {
         "static_current": "https://www.rejseplanen.info/labs/GTFS.zip"
       },
-      "operators": [
-        {
-          "onestop_id": "f-rejseplanen~dk~gtfs",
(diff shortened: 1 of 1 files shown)


drewda:
@akaise21 in this case I would recommend having no operator listing.

When the Transitland platform imports the feed, it will automatically detect the agencies and create "virtual" operator listings for all them.

It's only necessary to create operator listings in the Atlas when you want to add extra information (like a short name or tags) or to group togther multiple feeds (like one static GTFS and one GTFS Realtime feed for the same operator).

akaise21:
Thanks @drewda, got it! I've removed the operator in the newest commit :)

drewda:
thanks for your contribution!

akaise21:
Thanks for the fix ;)

PR #594, opened 2022-10-05
Update NORTA static feed

aquaticpotato3677:
No, "OpnesourceData" is not a typo. 
Access to feed requires creating an account and agreeing to their terms, which can be a bit finicky with their jank website, after which they email you a direct link with your key already appended to the URL.  Additionally, RT feeds are available, if you're willing to jump through a few more hoops, at the same URL.


drewda:
@aquaticpotato3677 thank you for this update

PR #595, opened 2022-10-05
Add River Valley Transit Authority

aquaticpotato3677:
Part of the Availtec MyBus tracking family. All agencies provide the same endpoints off of their base URLs, with static GTFS and GTFS-RT support. One example already in the database is the Akron Metro. Will update/add other agencies with this system soon.


drewda:
@aquaticpotato3677 thanks for this contribution.

We'll be glad to accept PRs for any other agencies as well. It's probably simplest to open a single PR for each agency. If you have a very large number, let us know and we can provide alternative means to manually provide a spreadsheet.

PR #605, opened 2022-10-10
update Rennes metropole GTFS

smevel:
update STAR source with new metro line


drewda:
@smevel thank you for this contribution

PR #622, opened 2022-10-14
update Transmilenio

aquaticpotato3677:
much bigger agency, might have made a mistake
source: https://www.arcgis.com/home/item.html?id=5cbd3053f4f546c6aafd3fd1c375e309
TIL you can search for arcgis data sets by GTFS tag and there's tons of data there


drewda:
@aquaticpotato3677 great find. Thanks! We usually try to stick with agency websites, but this is clearly created by a staffer for that agency.

drewda:
@aquaticpotato3677 FYI, just realizing after the fact that we should tag this feed with `tags: { "unstable_url": "true" }`. That is how we mark feed URLs with dates in them, or that will otherwise need to be manually checked from time to time.

PR #641, opened 2022-10-25
add gtfs for Sibiu (Tursib), Romania

szjozsef:


drewda:
@szjozsef I ran `transitland dmfr format --save` on this DMFR file. You can see the very small change that was applied.

Thanks for all contributions!

PR #646, opened 2022-10-31
add gtfs-rt

aquaticpotato3677:


aquaticpotato3677:
ok... upon further inspection, the agencies under Metrolinx have very odd things going on. Despite RT feeds specifying the agency (ex: https://metrolinx.tmix.se/gtfs-realtime-belleville/vehiclepositions.js), they return vehicles for all agencies listed [here](https://metrolinx.tmix.se/Tmix.Cap.DigSig.Gui/test/test.aspx). However, this doesn't seem to be the case for alerts and trip updates, as they return data for only the specified agency. Not quite sure how to handle this.

drewda:
@aquaticpotato3677 thank you for this attention to detail! The best way to model this is to probably create another RT feed record that only lists the vehicle positions endpoint — this RT feed record can then be referenced by all the operators. The agency-specific RT feed records can then just list the alert and trip update endpoints.

drewda:
looks good. thanks!

PR #672, opened 2022-11-16
update orange county url to add middletown

aquaticpotato3677:


drewda:
I see that both of those feed URLs are still listed on https://s3.amazonaws.com/datatools-511ny/public/index.html But neither of them have been updated in 2022.

We'll email the agency to check and see which to use. If you have any more info, feel free to share. Thanks.

drewda:
Both of these feeds are tracked at:

https://www.transit.land/feeds/f-newburgh~bus

https://www.transit.land/feeds/f-newburgh~beacon~shuttle~ny

PR #685, opened 2022-12-05
Update Link Transit Static GTFS Feed

brockhaywood:
This feed has been provided by Link Transit. 

Note: there is a conflicting PR for Link Transit here https://github.com/transitland/transitland-atlas/pull/675/files

However the feed version is newer in this PR, published Nov 17, 2022 whereas the trillium version in PR 675 is from Jan 28, 2022.


drewda:
@brockhaywood thanks for this submission. We heard from Trillium that they were no longer maintaining the source of truth for this feed, but had not yet been able to learn about its replacement.

I was having trouble editing this PR to also add matching RT feeds, so created a new PR that includes edits from this one: #686

PR #692, opened 2022-12-07
change hartransit to trillium link

aquaticpotato3677:


drewda:
Trillium produces this feed for Interline and one of our clients. We add in our license file and distribute it through the https://github.com/transitland/gtfs-archives-not-hosted-elsewhere repo, so that's the URL that third-parties should use when accessing the feed.

PR #693, opened 2022-12-07
Updated: Danville Mass Transit - f-danville~il~us

szjozsef:
Update source: https://www.ridedmt.org/services-programs/masstransit-advanced-components/search?q=gtfs


szjozsef:
It seems that the hosted feed is expired and does not get updated. Closing it.

PR #703, opened 2022-12-11
remove ceres area transit and stanislaus regional transit

aquaticpotato3677:
ceres and stanislaus have been merged (along with others) into one agency, so removing duplicated routes


drewda:
@aquaticpotato3677 great, thanks. It's always a bit of a challenge to figure out how and when agencies merge together.

PR #704, opened 2022-12-11
add hrt gtfs-rt

aquaticpotato3677:
unrelated to this particular pull but when the file was renamed in a prior pull, it appears to have broken the github link on the operator [page](https://www.transit.land/operators/o-dq9-hamptonroadstransithrt), where it still references the old file name


drewda:
@aquaticpotato3677 thanks for noticing. It appears that #553 successfully renamed the file field in the feed record but not in the operator record: 

image

We'll have a look.

drewda:
Should be fixed now:
image

PR #705, opened 2022-12-14
add link transit wa

aquaticpotato3677:


drewda:
@aquaticpotato3677 thanks. we recently added this in #686

aquaticpotato3677:
whoops! did remember something like that but didn't see it on the website in operators/feeds

PR #711, opened 2022-12-21
update: Transdev Royan Atlantique

szjozsef:
update source: https://transport.data.gouv.fr/datasets/donnees-gtfs-du-reseau-de-transport-public-cara-bus


drewda:
@szjozsef thank you for another useful update!

PR #742, opened 2023-02-06
Update with latest feed for bogota Colombia

mvanlaar:
Update with the latest gtfs feed from jan 2023


drewda:
@mvanlaar thank you for this update

PR #744, opened 2023-02-08
updated TEC data

fabrice1236:
Updated the TEC data source, since it was was from a third-party and outdated, to an official source. Also changed the website URL to their new one.


drewda:
@fabrice1236 thank you for this update

PR #753, opened 2023-02-20
Create dpmo.cz.dmfr.json

Peperman66:


Peperman66:
By the way, are the onestop id's good enough or should they be longer, including full names of the operator or feed provider?

drewda:
@Peperman66 thank you for this contribution. Those Onestop IDs look good. If you include the geohash component it is fine to have a short name. If you skip the geohash component, we aim to have a longer and more descriptive name.

drewda:
This will be ready to merge after its formatting is adjusted. To do so either run `transitland dmfr format --save feeds/dpmo.cz.dmfr.json` or mark the PR as ready for review and myself or another maintainer can apply the "opinionated format"

Peperman66:
I tried changing the spacing so now it should pass the tests

PR #765, opened 2023-02-28
Updated GTFS URL for ZET

franga2000:
The old URL was for a specific version of the feed and is no longer valid. This changes it to the always latest URL.   
 (see https://www.zet.hr/odredbe/datoteke-u-gtfs-formatu/669)

**EDIT:** looks like Github's editor forces newlines at the end of files, which the linter isn't happy about. I'll fix it once I'm back at a real computer.


drewda:
@franga2000 thanks for this contribution.

If you'd like a better web-based editing experience, try a GitHub Codespace: 

image

We have started using it with some other Transitland Atlas contributions, although still need to fully document the features.

PR #785, opened 2023-03-13
fix dl. URL for: CATS - Charlotte Area Transit System

szjozsef:
javascript - jquery redirect ....


drewda:
@szjozsef thanks. good catch!

PR #804, opened 2023-03-22
Update url for: Israel Public Transport

szjozsef:


drewda:
@szjozsef thanks for this update

Issue #807, opened 2023-03-28
why are some feeds (e.g. in Oregon) "historical"?


rosecitytransit:
By "historical" I mean that there's only a `static_historic` entry in `urls`

Specific example would be Sunset Empire Transportation District (`clatsopcounty-or-us`). According to https://oregon-gtfs.trilliumtransit.com/ the feed is currently valid.

I noticed that the routes weren't showing up on the Transit Land map.

Appears it may have been added in commit f4d73a6f30d294b6f688cb15319edb2ffa3caf5d by @evantrillium in #441

evantrillium:
@rosecitytransit this appears to be an error - there are lots of historical feeds published by Trillium for agencies that have either switched providers (eg, Corvallis), no longer offer services (eg, CEIC), or who have had their feeds materially reorganized (Sunset Empire/Clatsop used to be packaged with Tillamook and CCR.) The feed /clatsopcounty-or-us.zip *is* the most current one and should be used; apologies for the error.

Please let me know if you spot any others that you find suspicious and I'll have the team review.

rosecitytransit:
Can it be changed then in https://github.com/transitland/transitland-atlas/blob/master/feeds/oregon-gtfs.com.dmfr.json ? Also, if one of you have access to the https://oregon-gtfs.trilliumtransit.com/ Web page, the download link for Corvallis is missing.

PR #821, opened 2023-04-10
review Trillium feeds marked as historic

drewda:
re-activate, but retain `status=outdated` tag for tracking

related to #807

//cc @evantrillium


julieteldred:
@drewda is there anything else we need to handle on the Trillium end for the Oregon-GTFS feeds? Evan brought this ticket/#807 to my attention last week, but it looks like you have things under control

drewda:
@julieteldred I think we're up-to-date based on a quick look at Trillium's public list of feeds. But if there are more recent additions that we aren't covering, feel free to email me a list.

PR #831, opened 2023-04-19
update national park service gtfs feeds

eric-englin-volpe:
The listed feeds are now historic and a new feed has been added. The site has moved so I updated the file name as well.

### DIFF
--- a/feeds/nationalparkservice.dmfr.json
+++ b/feeds/nationalparkservice.dmfr.json
@@ -2,10 +2,24 @@
   "$schema": "https://dmfr.transit.land/json-schema/dmfr.schema-v0.4.1.json",
   "feeds": [
     {
+      "id": "f-9w8-grca~southrimshuttle",
+      "spec": "gtfs",
+      "urls": {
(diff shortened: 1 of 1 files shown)


drewda:
@eric-englin-volpe thanks for this contribution.

We did recently add the South Rim Shuttle feed to https://github.com/transitland/transitland-atlas/blob/master/feeds/nps.gov.dmfr.json

For the other feeds that you are marking as historic, is it just that there is no longer up-to-date GTFS data or those services are actually no longer operated?

eric-englin-volpe:
Oh interesting, thanks for adding the South Rim Shuttle Feed! 

And yes, those GTFS feeds are no longer up-to-date, but the transit services are still operating. We are working with the National Park Service to update those feeds and create new ones for 5-10 parks this summer. I was planning to update the nps dmfr as these new feeds are created. Let me know if another process may make more sense though.

drewda:
@eric-englin-volpe feel free to open new PRs whenever those other feeds are created. Alternatively, if they are released at the same existing URLs, Transitland will just fetch the update.

PR #834, opened 2023-04-20
remove wrong dart license

chadselph:
this license is for the wrong DART, (Des Moines Area Regional Transit Authority)


drewda:
@chadselph thanks for this fix

PR #835, opened 2023-04-21
fix readme broken header id link

sethidden:
Before this PR, clicking one of the links in the readme pointed to a `#onestopids` that doesn't exist. The ID is actually `#onestop-ids` (https://github.com/transitland/transitland-atlas#onestop-ids)


drewda:
@sethidden thanks for noticing and fixing this

PR #841, opened 2023-04-25
add GTFS Static feed for Białystok city in eastern Poland

sethidden:
I made up the Onestop IDs on the spot. Hopefully the json is ok :)


aquaticpotato3677:
two things: 
- onestop ids should use '\~' as the separator between names, with '-' only for separating the feed/operator prefix. so something like: f-bialostocka\~komunikacja\~miejska
- the linter also gets mad if you have a newline at the end of the file, so try removing that

aquaticpotato3677:
if it doesn't show a little red circle with a line through it at the end of the file, it thinks there's still a newline at the end of the file.

drewda:
@sethidden thanks for your contribution. And thanks to @aquaticpotato3677 for providing assistance.

PR #858, opened 2023-05-24
update url for: Irigo

szjozsef:
source: https://transport.data.gouv.fr/datasets/angers-loire-metropole-reseau-irigo-gtfs-gtfs-rt


drewda:
@szjozsef thank you for this update

PR #866, opened 2023-06-01
Update Feed - Link Transit, Burlington, NC

colbyjd3:
Updating the feed for Link Transit in Burlington, NC. New GTFS feed has its own domain, so a separate DMFR file has been created, and the previous feed data has been removed from transloc.com.dmfr.json (old feed has been added as historic on the new file).


aquaticpotato3677:
The linter doesn't like newlines at the end of files. You should see a little red circle with a cross through it if there's no newline.

colbyjd3:
> The linter doesn't like newlines at the end of files. You should see a little red circle with a cross through it if there's no newline.

That did it, thank you!

drewda:
Thank you both for this contribution!

PR #875, opened 2023-06-13
Update ridelbt.com.dmfr.json feed url

rileyqs:
Updating static_current and static_historic links to new and previous Long Beach Transit feed urls. The static_current feed url doesn't start until 6/18/2023 and the historic feed does not become historic until 6/18/2023.


drewda:
Thanks for this submission. FYI, we just made a similar update after receiving an email from Long Beach Transit staff: #878

PR #880, opened 2023-06-21
update some Gunma prefecture related providers

szjozsef:
The Gunma prefecture related GTFS files are no longer hosted at: https://gma.jcld.jp/GMA_OPENDATA/GTFS/...
They are in process to re-locate hosting for the GTFS files, some are hosted on the local provider websites some are hosted on one of the centrally hosted data platform sites (ex: https://gtfs-data.jp/, https://ckan.odpt.org/)

Source: https://www.pref.gunma.jp/page/198829.html

### DIFF
--- a/feeds/tshimada291.github.com.dmfr.json
+++ b/feeds/tshimada291.github.com.dmfr.json
@@ -2542,7 +2542,10 @@
       "id": "f-gtfs~1001~kanetsu~kotu~jp",
       "spec": "gtfs",
       "urls": {
-        "static_current": "https://gma.jcld.jp/GMA_OPENDATA/GTFS/gtfs.1001.kanetsu_kotu.zip"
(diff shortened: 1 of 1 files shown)


drewda:
The `feeds/tshimada291.github.com.dmfr.json` file is created from a list of Japanese feeds in https://github.com/tshimada291/gtfs-jp-list-datecheck

If possible, it would be best to update these feed URLs in that upstream repo. We can then re-run our script to rebuild the DMFR file: https://github.com/transitland/transitland-atlas/blob/master/scripts/convert-jp-list-csv-to-dmfr.py

If it's not possible to update the upstream repo, we can move those feeds to a different DMFR file and add them to the list of `existing_feeds_to_skip` in the script.

szjozsef:
Hopefully, the source list will be updated soon, see the issue there. This PR will be closed.

PR #883, opened 2023-06-22
Updates to Lisbon, Portugal feeds

ricardojorgerm:
Added the joint Carris Metropolitana area feed. Removed per-area (a2,a3,a4) Carris Metropolitana feeds. Removed old Lisbon metro area bus operators that were replaced by Carris Metropolitana. Removed discontinued Carristur airport bus brand.


drewda:
Thank you for helping to update these feed records

PR #903, opened 2023-07-25
Update lacmta.gitlab.com.dmfr.json

kylerchin:


kylerchin:
Thank you! Perhaps we can add a script into the github workflow that auto-fixes this?

I can make a PR

PR #926, opened 2023-08-20
Add CARRIS and change Carris Metropolitana history

joaodcp:
Firstly, CARRIS != Carris Metropolitana.

Lisbon (District) is composed of multiple municipalities. CARRIS only operates on one of those municipalities: Lisbon (Municipality) while Carris Metropolitana operates on some of the others (which includes Lisbon (Municipality), but limitedly because CARRIS is on an exclusivity agreement for that part of the district).

The first historical feed url for Carris Metropolitana was actually an old source for CARRIS feeds.

The url to the license I included applies to the old data that was provided via an intermediate, Transporlis.
I imagine the same license applies to the current data. CARRIS moved towards sharing the GTFS feed directly from their systems and do not provide any license info.

Thanks!

### DIFF
--- a/feeds/carris.pt.dmfr.json
+++ b/feeds/carris.pt.dmfr.json
@@ -26,5 +26,6 @@
                 }
             ]
         }
-    ]
+    ],
+    "license_spdx_identifier": "CDLA-Permissive-1.0"
 }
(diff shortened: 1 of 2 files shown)


drewda:
@joaodcp thank you for explaining the differences between these two systems. This is very helpful.

PR #947, opened 2023-09-22
Added feed for COTRAL S.p.A (Lazio, Italy)

jumpjack:


drewda:
Thanks for this submission. There is current an SSL certificate error when trying to download this feed. Transitland cannot currently fetch when there are any certificate issues. However, we are considering adding an override for this in the future. If we add that functionality, we will be able to merge this addition. Thanks for your patience.

jumpjack:
Which is exactly the issue? I can try contacting the webmaster in Italian.

Inviato da Posta per Windows

Da: Drew Dara-Abrams
Inviato: lunedì 25 settembre 2023 00:18
A: transitland/transitland-atlas
Cc: jumpjack; Author
Oggetto: Re: [transitland/transitland-atlas] Added feed for COTRAL S.p.A(Lazio, Italy) (PR #947)

Thanks for this submission. There is current an SSL certificate error when trying to download this feed. Transitland cannot currently fetch when there are any certificate issues. However, we are considering adding an override for this in the future. If we add that functionality, we will be able to merge this addition. Thanks for your patience.

jumpjack:
if the issue cannot be solved, can I temporarily load the GTFS on transit.land site, rather than downloading and installing the tool on my ststem?

Fale:
The issue is that the certificate is valid for `*.cotralspa.it` (`CN` field), while the domain is `travel.mob.cotralspa.it`. Due to how TLS certificate validation works, the initial star is only one level deep, so `travel.mob.cotralspa.it` does not match since it would be 2 levels deep. A correct certificate would be either `CN=travel.mob.cotralspa.it`, `CN=*.mob.cotralspa.it`, or `CN=*.*.cotralspa.it`

drewda:
@Altonss thank you for noticing this. Yes, it appears that it's now working locally when I use the transitland commands to download and validate the feed. This will hopefully now also work on the Transitland platform.

PR #953, opened 2023-09-25
update New York state feeds

drewda:
```sh
http https://s3.amazonaws.com/datatools-511ny/public/index.html | pup 'a attr{href}' | pbcopy
```


drewda:
Should also address https://github.com/transitland/transitland-atlas/pull/672

PR #956, opened 2023-09-27
update dc circulator url

wesleyschlenker:
Pulled from their developer resources: https://dccirculator.com/developer-resources/


drewda:
@wesleyschlenker thanks for these DC updates

PR #971, opened 2023-10-22
new url

piersoft:


piersoft:
new url gtfs

piersoft:
it's ok?

drewda:
@piersoft yes, thank you for adjusting this URL. It is unfortunate that they cannot provide a secure download via HTTPS, but it is fine for us to use HTTP.

PR #977, opened 2023-11-04
Update Polish GTFS sources

MKuranowski:
This PR comprehensively reworks listings for all GTFS and GTFS-RT feeds in Poland.

Apart from adding missing data, I have added comprehensive operator data (every feed links to at least one operator now). All IDs now also include a GeoHash.


drewda:
@MKuranowski thank you very much for so closely reviewing and expanding the coverage in Transitland Atlas for Polish feeds.

In cases of feeds and operators where a Onestop ID is already defined, we try to maintain this Onestop ID as long as it represents the same entity. This does mean that some Onestop IDs may lack geohashes or may have odd names. But we prioritize consistency over time so that data consumers can rely on the same IDs. Please revise this PR to keep the existing Onestop IDs.

Note that if there are specific situations where it does make sense to merge an existing feed record into another (or an operator record into another), there is a property called `supersedes_ids` that can take an array of Onestop IDs that have been merged into a given feed or operator.

MKuranowski:
Hi, thanks for the feedback.

I've made changes to stabilize the IDs. There are still some usages of `supersedes_ids`, let me go through the reasoning behind the ID changes:
- `f-u3q-ztmwarszawa~kolejemazowieckie~warszawskakolejdojazdowa` a feed with all 3 operators is no longer available, the data has been replaced by 3 new feeds (`f-u3q-warszawski~transport~publiczny`, `o-u3q-kolejemazowieckie` and `f-u3qb-warszawskakolejdojazdowa`
- `o-u3qc-komunikacjamiejskaŁomiankispzoo` was a duplicate entry for `o-u3qc-komunikacjamiejskaŁomianki`
- `f-u2zs-mpk~rzeszówspzoo` - feed is no longer available; `f-u2zs-rzeszow` is provided by a different entity
- `o-u3h4-mpkautobusy` and `o-u3h4-mpktramwaje` were duplicate entries for a single operator (`o-u3h4-mpk~wroclaw`)
- `o-u35g-sevibus~dlamiękinia`, `o-u3h4-dla~sevibussiechnice`, `o-u3h5-sevibus~dladługołęka` and `o-u3h55-dlawiszniamała` are all duplicate entries for a single operator (`o-u3h4-dla`).

drewda:
@MKuranowski thank you for providing these notes and for updating your PR. All makes sense and looks good.

drewda:
@MKuranowski for your information, the GitHub Actions job has identified two issues with this PR:

The opinionated DMFR format that is required within the Transitland Atlas repository does not currently allow `feed.urls.static_planned` fields. We've noted this bug for future follow-up at https://github.com/transitland/transitland-atlas/issues/979 For now, we'll merge this PR without those fields.

More than one feed record in the PR includes the URL `https://mkuran.pl/gtfs/wejherowo.zip` Can you please update the PR to remove one or merge them into one record?

MKuranowski:
Thanks, I've pushed a fix for that URL.

PR #981, opened 2023-11-21
Updated feed for I-RIDE Trolley.

jeffreydivi:
This PR updates the I-RIDE Trolley feed to be up to date.

The I-RIDE Trolley in the I-Drive Business Improvement District has switched from OmniModal (which is now redirecting to spam/ads) to BusWhere. This PR changes the I-RIDE Trolley to use the new vendor.

Data feed was sourced by staff at the I-Drive Business Improvement District. URL is `https://buswhere.com/iridetrolley/gtfs/schedule`.

The remaining OmniModal service, the [Sanford Community Redevelopment Agency trolley service](https://sanfordfl.gov/residents/free-trolley-service/), uses TSOMobile. I do not have their GTFS feed available; more research (i.e., contacting them) is necessary and is out of scope for this PR.

### DIFF
--- a/feeds/buswhere.com.dmfr.json
+++ b/feeds/buswhere.com.dmfr.json
@@ -1,0 +1,26 @@
+{
+  "$schema": "https://dmfr.transit.land/json-schema/dmfr.schema-v0.5.0.json",
+  "feeds": [
+    {
+      "id": "f-djn1-fl~iridetrolley",
+      "spec": "gtfs",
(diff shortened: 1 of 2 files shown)


jeffreydivi:
Converting to Draft for now because the data is stubbed for some reason. Contacting municipality.

drewda:
This feed still won't import, but at least it will fetch, so we'll merge this update

PR #997, opened 2024-01-15
gocary is moving their GTFS static feed

dradding:


drewda:
@dradding thanks for this update.

The file needs a few changes to its format to meet the "opinionated" DMFR format used in this repo. Please either allow editing by others to your branch and we will do it, or you can do it yourself by running:

```sh
transitland dmfr format --save feeds/gocary.org.dmfr.json
```

dradding:
Got it! I'll run that command on my end and push any updates

PR #1006, opened 2024-02-02
Jacksonville Transport Authority

Casual-Scrub:
New source for Jacksonville Transport Authority. Old one doesn't work.


drewda:
@Casual-Scrub thanks for this update!

This URL is not downloadable directly. Looks like the webpage at https://ride.jtafla.com/gtfs-archive may use some client-side JavaScript. We'll keep this PR open temporarily and look into other ways to access the download URL.

PR #1008, opened 2024-02-02
New source for Brampton Transit

Casual-Scrub:
The old URL for Brampton Transit wasn't working. I found the current webpage that GTFS Schedule feeds can be downloaded from.


drewda:
@Casual-Scrub thanks for this update

PR #1017, opened 2024-02-20
Update opentransportdata.swiss GTFS-feed "perma"-link

vkrause:


drewda:
@vkrause thank you for this update

PR #1023, opened 2024-03-02
It seems like they change url at every update :(

Fale:


drewda:
@Fale thanks for this contribution.

For this situation, we have an `unstable_url` tag. This will let our team and contributors search for just these feeds in the future: https://www.transit.land/feeds?tagUnstableUrl=true That way we can all make regular manual checks to see if the URL needs to be updated.

PR #1025, opened 2024-03-03
 Swap Red Deer historic and current feeds

Pi-Cla:
Older historic link fetches the latest feeds while the newer link has stopped working


drewda:
@Pi-Cla thanks for this contribution.

It does appear that https://data.reddeer.ca/gtfsdatafeed now points to the old/new URL

PR #1030, opened 2024-03-05
Update TPER links

Fale:


Fale:
What's the problem here? Incorrectly formatted is very vague.
Also, if I do the following:

```
$ transitland dmfr format feeds/solweb.tper.it.dmfr.json > /tmp/test
$ transitland dmfr lint /tmp/test | tail -1
2024-03-05T14:32:16Z [ERROR] Incorrectly formatted files: /tmp/test
```
So, even `format` creates invalid files without returning any error?

drewda:
@Fale I'm not sure why you are having that issue.

Try running `transitland dmfr format --save feeds/solweb.tper.it.dmfr.json`

Note that the difference with your earlier commit is the line break at the end of the file.

Fale:
Thanks! I was not aware of the `--save` option. I think the "weird" behavior on end-of-file newline could be related to the text editor config I have, which trims them, now that I know that dmrf is sensible to this, I will act accordingly :)

PR #1038, opened 2024-03-12
reverted VVTA's static to be from transloc

brockhaywood:
- confirmed with VVTA that the official static GTFS that is for public consumption is the transloc feed
Screenshot 2024-03-12 at 2 28 44 PM


drewda:
handled under #1044

PR #1041, opened 2024-03-13
Assign Realtime IDs to correct operators

kylerchin:


kylerchin:
It really ought to be easier to format transitland files without having to spend 2 hours attempting to compile glibc. 

```bash
transitland: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by transitland)
transitland: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by transitland)
transitland: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by transitland)
```

Maybe we can write a rust script to format it the way the linter wants it?

kylerchin:
I cannot run transitland for some reason

kylerchin:
Yay i did it!

drewda:
@kylerchin thanks very much for identifying these broken and mismatched IDs

Regarding the transitland CLI, pre-built binaries for some platforms can be downloaded from https://github.com/interline-io/transitland-lib/releases

PR #1050, opened 2024-03-16
Add invalid_tls_certificate tag

Fale:
With this change, I'm proposing to add such a tag.

I know that at the moment transitland has no use of it, but I think the starting point is adding the tag :).

Also, transitland-atlas is being used by many other project as a data repo, so I think this would still make sense even if for the foreseeable future transitland itself is not able to use it.

### DIFF
--- a/feeds/trentinotrasporti.it.dmfr.json
+++ b/feeds/trentinotrasporti.it.dmfr.json
@@ -12,6 +12,9 @@
         "url": "https://www.trentinotrasporti.it/open-data/",
         "use_without_attribution": "no",
         "create_derived_product": "yes"
+      },
+      "tags": {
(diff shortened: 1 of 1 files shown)


lolpro11:
I agree with this tag, although this tag should honestly be automated (no idea how)

Fale:
Automating it is risky because a certificate might be invalid for malicious reasons

lolpro11:
Good point. That was my only concern for this idea.

drewda:
Closing as out of scope. We've been hesitant to flag security issues like this in the public repo, but each user of Transitland Atlas is certainly welcome to keep their own additional metadata in side-car files that can be merged with the DMFR files, or used to control their software referenced against Onestop IDs.

PR #1053, opened 2024-03-21
Added Trois-Rivières and Drummondville

FelixINX:
Hello!

Here is Trois-Rivières and Drummondville.
Trois-Rivières has realtime data, I'll check with them if it's public.

Thanks!

### DIFF
--- a/feeds/drummondville.ca.dmfr.json
+++ b/feeds/drummondville.ca.dmfr.json
@@ -1,0 +1,20 @@
+{
+  "feeds": [
+    {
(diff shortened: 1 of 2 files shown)


drewda:
@FelixINX thanks for this contribution. We've been wanting to add Trois-Rivières for a long while now.

FelixINX:
@drewda Yes I'm happy that it's finally available!
There is a "Winter 2024" in the link, so I'm unsure if it will be updated. I will check with them and send a PR if needed.

PR #1059, opened 2024-03-27
Update data.bordeaux-metropole.fr.dmfr.json

kalon33:
Update feeds location (there's now only one feed for buses and tramways) and add realtime data.


drewda:
@kalon33 thanks for this update.

Note that when one feed is merged into another, we try to retain the old feed Onestop ID in `feed.supersedes_ids`

kalon33:
> @kalon33 thanks for this update.
> 
> Note that when one feed is merged into another, we try to retain the old feed Onestop ID in `feed.supersedes_ids`

Thanks @drewda , I knew for feeds URL but not for ids. I will try to remember this.

PR #1060, opened 2024-03-27
Create capcotentin.fr.dmfr.json

kalon33:
New GTFS and GTFS-RT feeds from Cap Cotentin. GTFS-RT feed contains vehicle_positions, trip_updates and realtime_alerts in the same feed.

### DIFF
--- a/feeds/data.gouv.fr.dmfr.json
+++ b/feeds/data.gouv.fr.dmfr.json
@@ -218,7 +218,7 @@
       },
       "tags": {
         "notes": "f-sp9x-normandie has a f-cap~cotentin subfeed that includes realtime data as the f-cap~cotentin~rt feed. Ids in the GTFS-RT only map to the static feed in f-cap~cotentin, not these in f-sp9x-normandie."
-      }
+      },
       "operators": [
         {
           "onestop_id": "o-aeroport~caen~carpiquet",


kalon33:
@drewda Could you have a look at this please? The `f-cap~cotentin~transit` feed is a subset of `f-sp9x-normandie` in data.gouv.fr.dmfr.json, and `f-cap~cotentin~transit~rt` is the RT feed linked to it. But I don't know how to reference it in data.gouv.fr.dmfr.json as it is indeed only relevant to a subset of `f-sp9x-normandie`.

Thanks for your help!

feeds/capcotentin.fr.dmfr.json
```diff
@@ -0,0 +1,43 @@
...
+        "url": "https://www.etalab.gouv.fr/licence-ouverte-open-licence/"
+      }
+    }
+  ],
+  "operators": [
+    {
+      "onestop_id": "o-cap~cotentin~transit",
+      "name": "Cap Cotentin",
+      "website": "https://www.capcotentin.fr",
+      "associated_feeds": [
```
drewda, line 32: Also associate this operator with `f-sp9x-normandie`

It's fine for an operator to group together multiple static feeds that can only be used exclusively.

feeds/capcotentin.fr.dmfr.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "https://dmfr.transit.land/json-schema/dmfr.schema-v0.5.0.json",
+  "feeds": [
+    {
+      "id": "f-cap~cotentin~transit",
```
drewda, line 5: Please add a `tags.notes` property on the feed describing the situation.

drewda: It would probably also be useful to add a `tags.notes` on to the `f-sp9x-normandie` record as well.

drewda: Finally, it would also be useful to add to this feed: `tags.exclude_from_global_query=true` The Transitland stack will then import both static feeds, but exclude the `f-cap~cotentin~transit` routes/stops/etc from global searches and tiles.

drewda (changes requested):

feeds/capcotentin.fr.dmfr.json
kalon33: I think I did what you requested but I also feel I messed up with the tags. I'm not sure I understood properly their structure. Could you have a look @drewda? Thanks for your review!

drewda: Thanks for all your edits. All looks good, other than that the `true` value needed to be in quotation marks.

drewda (approved):

kalon33:
Thanks @drewda!

PR #1061, opened 2024-03-27
Remove OCTA old alerts, Create City of irvine California

kylerchin:
OCTA has informed me that the alerts endpoint has been dropped since early 2023, over 11 months ago.

Irvine has a new bus service and we found the data urls for it.


drewda:
@kylerchin thanks for these contributions.

In the future, please open PRs from a branch with a name other than `main`. It makes it difficult to checkout another main from a different origin.

You can either apply the opinionated DMFR format to this PR. Or you can open a new PR from a differently named branch and I will apply the formatting.

PR #1069, opened 2024-03-30
Remove OCTA old alerts, Create City of irvine California 

kylerchin:
Duplicate of #1061 

OCTA has informed me that the alerts endpoint has been dropped since early 2023, over 11 months ago.

Irvine has a new bus service and we found the data urls for it.


kylerchin:
Thank you for the information @drewda ! Perhaps it can be placed in the readme that PRs should use seperate branches if the contributor needs assistance formatting? I for some reason cannot get transitland-lib running on WSL as `glibc` is not working.

I was thinking about writing a quick duplicate of the formatter + linter in rust so it could be installed via cargo.

drewda:
Sure, we can add mention of this in the readme. Note that this will an issue with any public GitHub repo -- it's normally best to open all PRs from branches with other names.

Not sure how to best run `transitland-lib` under Windows Subsystem for Linux. But other options that definitely do work are:

- start a GitHub Codespace against this repo: https://docs.github.com/en/codespaces
- run the devcontainer in this repo on a local copy of Docker. Visual Studio Code can assist with this: https://code.visualstudio.com/docs/devcontainers/containers

PR #1071, opened 2024-04-05
New Feeds - Salisbury Transit & Gastonia Transit

colbyjd3:
Adding new feeds for Salisbury Transit in Salisbury, NC, and Gastonia Transit in Gastonia, NC, to arcgis.com.dmfr.json.


colbyjd3:
Linter seems to be unhappy with one of my feed IDs, even though the other follows the same format. Would appreciate if someone could point out what I got wrong; I'm trying to get better at this. Thank you!

PR #1075, opened 2024-04-07
Update failing feed URLs for some transit providers in Virginia

mwootten:
A few feeds have been failing since their URLs changed:
* Fairfax City CUE changed their (unstable) feed URL to another page on their website
* Danville Transit and Pulaski Area Transit both migrated to a new provider


drewda:
@mwootten thanks for these updates

PR #1081, opened 2024-04-10
update the BFT feed source of truth per BFT team

brockhaywood:


drewda:
@brockhaywood thanks for this update.

Do you know if this agency also happens to have RT endpoints? If so, it would be great if we could add those.

PR #1083, opened 2024-04-12
Update - Multiple Feeds for GoCary, GoDurham, GoRaleigh, and GoTriangle

colbyjd3:
Per [https://gotriangle.org/developer-resources](https://gotriangle.org/developer-resources), each of these operators has at least two active GTFS feeds: one hosted on Trillium, and the other hosted on their local domains (e.g. `gotriangle.org`). By comparing these feeds on [https://gtfs-validator.mobilitydata.org/](https://gtfs-validator.mobilitydata.org/), it seems that the Trillium feeds generally provide more data than the ones hosted on local domains despite being marked as `static_historic` on each *.dmfr.json file. As such, I moved the Trillium URLs over to `feeds/trilliumtransit.com.dmfr.json` and marked those new feeds as `associated_feeds` under `operators`. Lastly, I added realtime feeds for GoCary, GoRaleigh, and GoTriangle.

### DIFF
--- a/feeds/passio3.com.dmfr.json
+++ b/feeds/passio3.com.dmfr.json
@@ -378,6 +378,27 @@
       ]
     },
     {
+      "id": "f-dq259-ncsu~wolfline~nc~us",
+      "spec": "gtfs",
+      "urls": {
+        "static_current": "https://passio3.com/ncstateuni/passioTransit/gtfs/google_transit.zip",
+        "static_historic": [
+          "https://api.transloc.com/gtfs/ncsu.zip"
+        ]
+      },
+      "operators": [
+        {
+          "onestop_id": "o-dq259-ncsuwolfline",
+          "name": "NCSU Wolfline",
+          "tags": {
+            "twitter_general": "NCStateTranspo",
+            "us_ntd_id": "40147",
+            "wikidata_id": "Q8030170"
+          }
+        }
+      ]
+    },
+    {
       "id": "f-eastern~virginia~medical~school",
       "spec": "gtfs",
       "urls": {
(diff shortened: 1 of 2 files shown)


colbyjd3:
Linter is still unhappy with the formatting for `gocary.org.dmfr.json` and `gotriangle.org.dmfr.json`. I've been comparing them to the working .json files and am running out of ideas. Will revisit tomorrow.

drewda:
@colbyjd3 thanks for these contributions.

We rarely add multiple static feeds for the same agency. Ideally we'd hear from the transit agencies on why it's useful to have multiple overlapping feeds. Alternatively, if you don't have a contact at those agencies, please most more information on the stops/routes that are only available in the Trillium-created static feed and how that feed's IDs compare with the static feeds produced by the RT vendors' systems. Thanks.

colbyjd3:
Thanks @drewda, that makes sense. It does appear that, of the static feeds, the Trillium ones contain the most data. I may rework this to set the Trillium feeds as primary for static and simply reserve the agency domain .json files for the RT feeds. That said, a couple of these agencies (GoDurham and GoRaleigh) are implementing service changes over the next couple of weeks, so I may keep an eye on that and see which is updated first (I'd assume both, but it's not a bad idea to keep tabs before switching the primary static feed).

As always, greatly appreciate your input and patience. Thank you!

brockhaywood:
my preference would always be to use the source of truth for these feeds which i believe to be on the agency domains (in this case). in the past i have found that the trilliium ids don't match those from the source of truth which can be problematic when correlating other data from the agency.

colbyjd3:
Thanks @brockhaywood, I have made the recommended adjustments via Pull #1105. Greatly appreciate both your and @drewda's input on this!

PR #1085, opened 2024-04-18
Update transcollines.github.dmfr.json

brodyFlannigan:
Change URL to use the Fares-V2 dataset

Interline peeps: the RT URLs have been discontinued for quite some time while Transcollines upgrades to a new CAD/AVL system. I'm not sure how best to go about removing these links. Any assistance would be greatly appreciated.


drewda:
@brodyFlannigan thanks for this update.

We'll remove the RT feed record for the time being. Please do open a PR if/when there's a new RT feed or drop us a line at hello@transit.land

PR #1095, opened 2024-04-25
Update carrismetropolitana with RT feeds

ricardojorgerm:
This change adds the realtime feeds from Carris Metropolitana, and additional metadata (WikiData, short name, twitter).


drewda:
@ricardojorgerm thank you for this contribution

ricardojorgerm:
Your live webpage is mixing the onestop ids of carris metropolitana and carris, is that something we can fix here?

PR #1102, opened 2024-05-02
Update metrotas.com.au.dmfr.json

martiandeath:
Added two additional URLs for Metro Tasmania as they have three current URLs for different parts of the state (the added URLs also include new operators Lee's Coaches, Manion's Coaches, Calow's Coaches, East Tamar Bus Lines and Mersey Link, as well as Redline Coaches and Tassielink, which were already in the exisiting URL)

### DIFF
--- a/feeds/metrotas.com.au.dmfr.json
+++ b/feeds/metrotas.com.au.dmfr.json
@@ -2,22 +2,46 @@
   "$schema": "https://dmfr.transit.land/json-schema/dmfr.schema-v0.5.0.json",
   "feeds": [
     {
-      "id": "f-r22-metrotasmania",
+      "id": "f-south-metrotasmania",
       "spec": "gtfs",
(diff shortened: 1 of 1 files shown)


drewda:
@martiandeath thanks for this contribution.

The DMFR format cannot take more than one current static URL for a single feed record. The pattern to use instead is to create a separate feed record (with a separate feed Onestop ID) for each one of the current static feed URLs.

martiandeath:
Good to know, I'll see if I can figure that out later today, cheers

martiandeath:
@drewda I think I have modified it correctly this time, let me know if something is wrong

drewda:
@martiandeath thanks for revising!

I've made a few edits to retain existing Onestop IDs (to be more stable for Transitland's users).

Will go ahead and merge this now.

PR #1103, opened 2024-05-04
Disambiguate Carris and Carris Metropolitana

ricardojorgerm:
Referring to a different reporter's explanation in: https://github.com/transitland/transitland-atlas/pull/926#issue-1858236772

These are two different operators.

Currently, there are two onestop ids in the website:

o-eyc-carrismetropolitana
o-eyck-carris

The latter currently contains both Carris Metropolitana and Carris, which is wrong.
In this update, I've aligned the onestop IDs present on the website with the onestop IDs in the files, by associating:

o-eyc-carrismetropolitana -> Carris Metropolitana
o-eyck-carris -> Carris

Additionally, I've been told that Carris Metropolitana is now suggesting all users to refer to the API for all feeds, not just RT ones, so I put the new URL in.

### DIFF
--- a/feeds/carris.pt.dmfr.json
+++ b/feeds/carris.pt.dmfr.json
@@ -12,14 +12,18 @@
       },
       "license": {
         "url": "https://dadosabertos.cm-lisboa.pt/dataset/informacao-sobre-transportes-publicos-da-cidade-de-lisboa-carris"
-      }
-    }
-  ],
-  "operators": [
-    {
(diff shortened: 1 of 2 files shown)


drewda:
@ricardojorgerm thanks for explaining the differences between the two agencies, and also for updating one of the feed URLs.

The main problem was that the o-eyckr-carris operator was not actually associated with the f-carris~pt feed. And therefor a "virtual" operator record was being created.

I've fixed this and made a few other additions.

PR #1105, opened 2024-05-12
Update - Add Live Feeds and Other Attributes for Research Triangle Operators

colbyjd3:
Live feeds and additional attributes have been added for GoRaleigh, GoDurham, and GoCary. Additionally, new files have been created for GoTriangle and Chapel Hill Transit, and previous feeds for each have been removed from the Trillium Transit file.

Note that this is a reattempt of Pull #1083, which had fallen far behind on commits and could no longer be utilized.


colbyjd3:
This is probably the fourth or fifth time I've encountered this exact issue with formatting. In previous encounters with this problem, I believe the issue was primarily due to my default tab spacing, but I've adjusted it to two spaces, which seems to be the default for other DMFR files. I'm not seeing any trailing spaces this time either (went line-by-line). And, when I look at the failed job, nothing seems to be highlighted:

image

I hate that y'all have to keep fixing my formatting errors, so I have to ask: is there a way to drill down further into the error report and find the exact location of said error? And, if not, what are some common places that I can check first in terms of formatting? Thank you!

colbyjd3:
Finally managed to get transitland-lib configured in VSCode on my local machine, ran opinionated DMFR format, and got all checks to clear. This one should be ready!

PR #1108, opened 2024-05-14
Add feed: COLT in Loveland, Colorado

markstos:
My first pass a contributing a feed. Feedback welcome.

Our team is able to download this feed to our laptops, but we get a 403 when running the same download command from our servers. I'll be interested to see if your automation gets a better response.


drewda:
@markstos thanks for this contribution

PR #1114, opened 2024-05-23
Bump Translink Vancouver GTFS-rt to v3

kylerchin:
New version documentation published here: https://www.translink.ca/about-us/doing-business-with-translink/app-developer-resources/gtfs/gtfs-realtime

### DIFF
--- a/feeds/translink.ca.dmfr.json
+++ b/feeds/translink.ca.dmfr.json
@@ -42,9 +42,9 @@
       "id": "f-translink~rt",
       "spec": "gtfs-rt",
       "urls": {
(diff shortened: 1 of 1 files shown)


kylerchin:
Can we remove this requirement, it's quite annoying to deal with, and the dmfr linter provided does not compile on my machine for some reason.

drewda:
@kylerchin we're sorry to hear your frustrations. We are not making any changes at this point but will note this for future consideration.

Not sure how you're currently editing the DMFR files, but switching to a code editor will give you more control. An editor like vi or Visual Studio Code typically won't modify the line break at the end of a file when you're only saving changes to other lines in that file. You can also use a code editor to show invisible characters and to see what type of line break characters are being inserted.

PR #1122, opened 2024-06-06
Update - GoRaleigh (Raleigh, NC, USA)

colbyjd3:
Correcting `static_current` in an effort to pull new system route changes. Also adding license attribution and shifting `operators` outside of `feeds`.


drewda:
@colbyjd3 thanks very much for this targeted update

PR #1124, opened 2024-06-07
Update static URL in data.bloomington.in.gov.dmfr.json

NomeQ:
@drewda It looks as if this new URL may also be stable. Should I remove the unstable URL tag as well?


drewda:
@NomeQ yes, you can go ahead and remove the unstable_url tag

PR #1126, opened 2024-06-07
Update static URL for dot.nm.gov.dmfr.json

NomeQ:


drewda:
@NomeQ looks good. Thanks!

PR #1127, opened 2024-06-08
Added Mwasalat GTFS and RT-GTFS

AlkindiX:


drewda:
@AlkindiX thanks for this contribution

PR #1128, opened 2024-06-10
Bloomington Transit changed their URL again.

markstos:


drewda:
@markstos thanks for this contribution.

Note that we just updated this feed URL in https://github.com/transitland/transitland-atlas/pull/1124 Hopefully that URL will be more stable going forward

PR #1145, opened 2024-06-14
Add new Comboios de Portugal (CP) feed

joaodcp:
Hi,

This adds the new Comboios de Portugal feed that is now provided directly by them instead of being served via Transporlis.
I also removed the feed from the Transporlis DMFR file.

Thanks! 🚀


drewda:
@joaodcp thank you for this update

PR #1161, opened 2024-06-27
Remove City of Avalon fixed-route service

triphora:
Was completely replaced with microtransit in April.


drewda:
@triphora thanks for reporting this change!

We'll use a different PR to mark this static feed as archived.

PR #1177, opened 2024-07-29
Update mct.org.dmfr.json

hvents4:
added the current gtfs data link


drewda:
@hvents4 thank you for this update

PR #1178, opened 2024-07-29
Update metrarail.com.dmfr.json

hvents4:
updating static gtfs link -- old one didnt work


drewda:
Thank you for this PR.

The URL at https://gtfsapi.metrarail.com/gtfs/raw/schedule.zip does still work. It just requires authentication, per https://github.com/transitland/transitland-atlas/blob/10eceba12809ec78372118c1cb930161b449e9e5/feeds/metrarail.com.dmfr.json#L15-L18

PR #1193, opened 2024-08-09
Add feed file for daytrip gtfs feeds

joywang8:
We are Daytrip Shuttle, and we operate multiple shuttle routes around the world.
We would like to add first 2 feeds in Italy and Portugal in Transit.Land and more feeds will be coming.


drewda:
@joywang8 thanks for these contributions. I've adjusted the format slightly and will merge

PR #1228, opened 2024-08-30
Update data.gouv.fr.dmfr.json : feed URL change

kalon33:
Fix f-sp9x-normandie feed URL which changed recently.


kalon33:
@drewda this is only a URL fix to use the new one as the previous was 404'ing for a while.

Thanks!

PR #1257, opened 2024-10-08
Regenerate Toscana dati from feed

joeyates:
This PR replaces the existing data for Tuscany with data taken from [OpenToscana's official feed](https://dati.toscana.it/dataset/rt-oraritb.xml).

Major changes:
* A number of bus companies in the previous file no longer exist and have been removed,
* All Tuscan bus routes are now covered,
* URLs are now all https,
* Geohashes are calculated to be as precise as possible while covering all stop locations.

The file is generated via this repo: https://github.com/joeyates/transitland-dati-toscana-it

### DIFF
--- a/feeds/dati.piemonte.it.dmfr.json
+++ b/feeds/dati.piemonte.it.dmfr.json
@@ -787,19 +787,7 @@
         "use_without_attribution": "no",
         "create_derived_product": "yes",
         "attribution_text": "Elaborazione  dati  realizzata  da  [Licensee],  basata  su  \"Servizio programmato del Trasporto Pubblico Regione Piemonte - Treni Regionali\"  della  Regione Piemonte"
-      },
-      "operators": [
-        {
-          "onestop_id": "o-u0j-trenitalia",
-          "name": "TRENITALIA",
-          "website": "http://www.trenitalia.it",
-          "associated_feeds": [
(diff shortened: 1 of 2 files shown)


drewda:
@joeyates thank you very much for all these updates.

Are any of the newly added feeds replacements for removed feeds? If so, please associate the old Onestop IDs using a `supersedes_ids` property.

Here is an example of how multiple outdated feeds were replaced by a single current feed record: https://github.com/transitland/transitland-atlas/blob/b3c3de9716d90c09420ecfcc4edff9d4d2075522/feeds/carrismetropolitana.github.com.dmfr.json#L6-L13

And here is an example of how an two operators were merged into one: https://github.com/transitland/transitland-atlas/blob/b3c3de9716d90c09420ecfcc4edff9d4d2075522/feeds/carrismetropolitana.github.com.dmfr.json#L41-L43

Thanks!

joeyates:
Hi @drewda 

I've added `supersedes_ids` where possible.

I've added back a couple of feeds that may be obsolete, but I think it's better that they stay in place until a new feed is available.

drewda:
Hi @joeyates great, thank you for your additions.

If any of these feeds become active again at new sources, we'll look forward to additional updates. Thanks.

PR #1266, opened 2024-10-16
Update bow valley from trillium to consat

dradding:
Please scrutinize this heavily. I'm not sure if I've done conversion correctly. Specifically, their IDs seem to predate the current conventions.


drewda:
@dradding thanks for this update.

You're right that one of these feed Onestop ID is from a very old process. It looks ugly, but in general we try to retain Onestop IDs unless it's really necessary to change them. I'll tweak that and then merge.

PR #1268, opened 2024-10-21
Update the mvv's url

CommanderStorm:
This fixes the MVVs URL being a 404.

This will need to be updated semi-regularly, to their new arbitrairy timestamp :sigh:

A bit older, but with shapes data exists here:
https://www.govdata.de/suche/daten/soll-fahrplandaten-mvv

Sadly, this is not listed on their website
=> idk what the timestamp is for the ones with shapes


drewda:
@CommanderStorm thanks for this update

PR #1319, opened 2024-11-01
Change STCP GTFS URL to 2024 version

joaodcp:
Updated the url.


drewda:
@joaodcp thank you for this update

PR #1323, opened 2024-11-05
Additional Feeds for NC Agencies

colbyjd3:
New agencies and feeds for Iredell County Area Transportation System, Goldsboro-Wayne Transportation Authority, and Greenville Area Transit in North Carolina.

~Updating feed for Salisbury Transit.~

~Lastly, adding second feed for Sunway Charters, which operates [three additional routes](https://sunwaycharters.com/bus-tickets/) that are separate from the Amtrak Thruway program. (Note: I'm not sure how you want to handle this one, as we generally don't do multiple static feeds for a single operator. I've tried keeping the operator block in `feeds\amtrak.com.dmfr.json` and moving it to `feeds\gtfs.transitportal.org.dmfr.json`, but it doesn't seem to like either option.)~

It is worth noting that the new NTD GTFS Weblinks file does not list official feeds for any of these operators and that gtfs.transitportal.org appears to be a project of NC State's Institute for Transportation Research and Education. While some North Carolina agencies opted to utilize this project site for their NTD feed submissions, these ~five~ three (among others) did not.

### DIFF
--- a/feeds/amtrak.com.dmfr.json
+++ b/feeds/amtrak.com.dmfr.json
@@ -18,6 +18,9 @@
           "associated_feeds": [
             {
               "gtfs_agency_id": "51"
+            },
+            {
+              "feed_onestop_id": "f-amtrakvermonter~vt~us"
             }
           ],
           "tags": {
--- a/feeds/amtrak.com.dmfr.json
+++ b/feeds/amtrak.com.dmfr.json
@@ -154,16 +154,6 @@
           ]
         },
         {
-          "onestop_id": "o-dq0-sunwaycharters",
-          "name": "Sunway Charters",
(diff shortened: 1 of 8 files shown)


drewda:
Thanks for finding and reviewing gtfs.transitportal.org

https://gtfs.transitportal.org/CurrentGTFS/Sunway_Trolleys/Trolleys_20221105.zip and https://gtfs.transitportal.org/CurrentGTFS/Salisbury%20Transit_237/Salisbury_20190924.zip look too old to be worth adding as current feed URLs.

Unless you can see other sources for those feeds, let's just skip them for now.

colbyjd3:
Thanks @drewda, this is done!

PR #1326, opened 2024-11-08
Add 6 agencies

NomeQ:
Add batch of Trillium feeds, including flex feeds for existing feeds.
- eastbywest-nz
- eastside-wa-us
- eccog-co-us
- elpaso-co-us
- grandvalleytransit-flex
- gunnisonvalleyrta-co-us-flex


drewda:
@NomeQ all looks good!

PR #1342, opened 2024-11-24
Update Feed - Winston-Salem Transit Authority

colbyjd3:
WSTA in Winston-Salem, NC, has finally published a GTFS permalink on their website. Updating `static_current` to the new permalink, and renaming the DMFR to match the new URL.


drewda:
@colbyjd3 great, thanks for this update

PR #1347, opened 2024-12-05
Médio Tejo (Portugal)

claudioap:
This is the official feed for Médio Tejo.

We don't have a better (non-Google Drive) URL at hand.  This is the very same URL that sources Google Maps.

I do have the authorization to publish it under this license.

Cheers,
Cláudio


drewda:
@claudioap thank you for these contributions

PR #1348, opened 2024-12-05
CIM Leiria / Rodoviária do Lis (Portugal)

claudioap:
This is the official feed for CIM Leiria / Rodoviária do Lis.

We don't have a better (non-Google Drive) URL at hand.  This is the very same URL that sources Google Maps.

I do have the authorization to publish it under this license.

The branding of the operation is "Rodoviária do Lis"; perhaps we should have that as an explicit feed name; what do you think? 

Cheers,
Cláudio


drewda:
@claudioap a belated thanks for your contributions

PR #1358, opened 2024-12-19
Fix spelling mistake in Toronto Transit Commission

toofishes:
This seems to trickle through all the way to https://www.transit.land/operators/o-dpz8-ttc


drewda:
@toofishes thanks for noticing

Issue #1034, opened 2024-03-09
Add Koleje Dolnośląskie (Poland)


etam:
See https://kolejedolnoslaskie.pl/rozklady-gtfs/

drewda:
> The GTFS RT feed is also available but the site asks to e-mail for access. I assume it'll be password protected. How does transitland handle storing the passwords?

For the stack that runs transit.land, the Interline team maintains an internal list of credentials that are mapped onto individual feeds by Onestop ID or onto all of the feeds in a given DMFR file.

Other users of the Transitland Atlas repo may implement their own means of using their own credentials.

PR #1360, opened 2024-12-26
Add realtime_trip_updates for kolejedolnoslaskie

sethidden:
I requested for GTFS creds for KD in https://github.com/transitland/transitland-atlas/issues/1034 so now I'm on their mailing list.

I' got an e-mail saying that the /vehicles endpoint provided GTFS-RT VehiclePositions, but now they have a /feed endpoint where they provide both VehiclePositions and TripUpdates

They explicitly ask to *switch* from /vehicles to /feed


drewda:
@sethidden yes, you're right that it's fine if a given RT endpoint includes all of the RT message types

PR #1364, opened 2025-01-01
fix: metrarail gtfs url updated and authorization removed

HelloJowet:
In the [developer docs](https://metra.com/developers) they are recommending to use a different url for the gtfs feed that isn't secured by any type of authentication


drewda:
Thanks for this contribution. As Transitland is also sourcing real-time feeds from Metra, we want to use the same host and endpoints for both static and real-time.

PR #1365, opened 2025-01-01
fix: authorization in tcl feed removed

HelloJowet:
No authorization is needed to download this feed and no authorization is mentioned in the [official documentation](https://transport.data.gouv.fr/resources/65812?locale=en)


drewda:
Yes, basic auth is needed to download from that URL and it is documented at https://data.grandlyon.com/portail/fr/connexion

PR #1366, opened 2025-01-04
Update static gtfs feed link of Rhein-Neckar-Verkehr GmbH

networkException:
See https://www.opendata-oepnv.de//ht/de/organisation/verkehrsunternehmen/rnv/openrnv/datensaetze?tx_vrrkit_view%5Bdataset_name%5D=soll-fahrplandaten-rnv&tx_vrrkit_view%5Baction%5D=details&tx_vrrkit_view%5Bcontroller%5D=View


networkException:
Note that the previous link is broken so I haven't added it to `static_historic`. Technically RNV also offers historic versions of their feed, I don't think its necessary to maintain a directory of those here though.

drewda:
@networkException thanks for this contribution.

Note that we do want to retain previous URLs to be able to understand how they change over time

PR #1368, opened 2025-01-08
Update Prince Albert feed url [CA]

Altonss:


drewda:
@Altonss thank you for this update

PR #1375, opened 2025-01-14
Hamilton city feed url update

BaptistePignier:
Since the city of Hamilton updates its feed 4 times a year, I left the old URL as is.


drewda:
@BaptistePignier thanks for this contribution

PR #1390, opened 2025-01-15
Updated Chemung County C TRAN to use all caps in agency name

BNewborn:
I believe `static_current` for Chemung C Tran is out-of-date using lower-case for its url. On [NY State's 511 page](https://511ny.org/developers/resources), Chemung is linked to an all caps URL, but Transitland is using an older feed that has lower-case in its name.
image

The current lower-case feed has a calendar.txt that ends with `20221231` (left, in the picture below), while the upper-case feed (available on the 511 site above, right in the picture below) has feeds that go from 2023 to 2025
image

I also added the old URL to `static_historic`, as per the instructions on the main repo README.md


drewda:
@BNewborn thanks for this update

Issue #1369, opened 2025-01-08
Handling GTFS with Non-Static URLs for ZTM TransportGZM


grzesjam:
Hello! 👋 

I need some help/guidance regarding how this project would prefer to handle GTFS data when the URL for the static GTFS file dynamically changes.

**Context:**
ZTM TransportGZM is a public transportation organization based in the Silesian Voivodeship, Poland. They manage buses, trams, and trolleybuses. The organization provides both GTFS-RT and GTFS datasets:
- The **GTFS-RT** has a static URL, so setting it in `dmfr.json` file is straightforward. Licensed under CC-BY-4.0
- However, the **GTFS data** does **not** have a static URL. Instead, a new GTFS zip file is uploaded every few hours to their CKAN instance (dataset management software). Licensed under CC-BY-4.0.

The dataset can be accessed at:  
https://otwartedane.metropoliagzm.pl/dataset/rozklady-jazdy-i-lokalizacja-przystankow-gtfs-wersja-rozszerzona

Alternatively, there’s an API endpoint:  
`https://otwartedane.metropoliagzm.pl/api/3/action/package_show?id=317435cc-0075-4d10-b8ef-6e9b0010e90a`

Using the API, the newest GTFS file URL can be fetched dynamically with:  
```bash
jq -r '.result.resources | sort_by(.created) | last | .url'
```

**My Question:**
What approach would best align with this repository and its downstream dependencies (such as [transitous](https://github.com/public-transport/transitous))? 

1. Should I implement a GitHub Action in this repository to update the GTFS URL dynamically, similar to other projects? (GBFS, GBFS-jp)
2. Alternatively, should I create a separate repository with a GitHub Action that releases the GTFS under a static URL? (Although this would essentially make my repository act as middleware for the data.)

I would appreciate any advice or suggestions on how to proceed. Thanks in advance! 🙏

grzesjam:
Hello @drewda 
Can you help?

drewda:
Hi @grzesjam thanks for your contribution and for your patience.

We have a few different ways of handling them:

1) Looking more closely for ways to get a stable URL. I see that this is CKAN. Maybe there is a hidden CKAN endpoint to get the latest version? You've likely already considered this, but just mentioning it in case you haven't

[...]

PR #1403, opened 2025-01-18
Add PL Silesian buses (Transport GZM)

grzesjam:
Hello! 👋

`ZTM TransportGZM` is a public transport organization based in the Silesian Voivodeship, Poland. They manage buses, trams, trolleybuses and bikes. 
ZTM is the organization that governs it, "[TransportGZM](https://transportgzm.pl)" is a brand under which they run busses/trams/trolleybuses, and brand "[TransportGZM MetroRower](https://metrorower.transportgzm.pl/)" is for bikes. That's the reason I picked `o-transport~gzm~pl` and `f-transport~gzm~pl`/`o-transport~gzm~pl~tf`

This PR adds ZTM TransportGZM GTFS and GTFS-RT.
- The GTFS-RT has a static URL. Licensed under CC-BY-4.0
- The GTFS doesn't have a static URL. I [created an issue here](https://github.com/transitland/transitland-atlas/issues/1369) but got no response, so I just went with creating a mirror for their data. Data is licensed under CC-BY-4.0

If something is missing or please let me know :)

### DIFF
--- a/feeds/transportgzm.pl.dmfr.json
+++ b/feeds/transportgzm.pl.dmfr.json
@@ -1,0 +1,47 @@
+{
+  "$schema": "https://dmfr.transit.land/json-schema/dmfr.schema-v0.5.0.json",
+  "feeds": [
+    {
+      "id": "f-transport~gzm~pl",
+      "name": "ZTM Transport GZM",
+      "spec": "gtfs",
+      "urls": {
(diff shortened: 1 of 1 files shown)


grzesjam:
TransportGZM has two versions of GTFS: "Normal" (apparently "temporary") and "Extended"
Data inside of them different, it has different fields in files.

I tried to push this directly to `transitous` but it had issue parsing "Extended" version and i had to use "normal" one. 
I can't find anything about these extended and non-extended versions, which one should be used? 

Both version can be checked out in Org i created with mirror https://github.com/orgs/TransportGZM-GTFS-mirror/repositories

PR #1416, opened 2025-01-29
Update wvu.edu.dmfr.json

Altonss:
Update WVU feed: https://prt.wvu.edu/regulatory-compliance/gtfs


drewda:
@Altonss thank you for this update!

PR #1421, opened 2025-01-30
fix the newly changed format workflow

drewda:
- shouldn't have run on main: https://github.com/transitland/transitland-atlas/actions/runs/13058447897

- didn't work on PR on branch in the repo, as opposed to PR from a fork: https://github.com/transitland/transitland-atlas/actions/runs/13058213324


drewda:
note that the runs in the PR aren't actually testing the new workflow, since it's using the workflow file from main

PR #1424, opened 2025-01-31
new england feeds

aquaticpotato3677:
various new england feeds, most from ntd

### DIFF
--- a/feeds/nationalrtap.org.dmfr.json
+++ b/feeds/nationalrtap.org.dmfr.json
@@ -175,6 +175,23 @@
       ]
     },
     {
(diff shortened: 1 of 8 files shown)


aquaticpotato3677:
fails because i deleted some moved agencies from `2023-gtfs-weblinks.dmfr.json` as according to the readme

drewda:
@aquaticpotato3677 thanks for these additions

Hadn't thought about how the security check for changes under the scripts directory affects this, so I'll find a better place for `2023-gtfs-weblinks.dmfr.json` in a future PR

PR #1440, opened 2025-02-12
Add horizons.govt.nz - NZ Horizons Regional Council

Hoverth:
Hello!

After adding HRC to transitous in <https://github.com/public-transport/transitous/commit/842e347447249060844cdaf8fe0f865064beff4d>, I thought I'd take a stab at adding it here as well :)

I have tried to do everthing in accordance with <https://www.transit.land/documentation/atlas>, but feel free to give any pointers, as this is my first time contributing to transitland-atlas.


drewda:
@Hoverth thank you for contributing

PR #1444, opened 2025-02-20
Charm City Circulator - Baltimore, MD

BNewborn:
The current Charm City Circulator (Baltimore, MD) [feed](https://www.transit.land/operators/o-dqcx8-charmcitycirculator#routes) only has 4 routes - it is missing [the new Cherry Route](https://baltimorebeat.com/charm-city-circulator-adds-cherry-hill-route-connecting-the-neighborhood-to-downtown/).

I found a feed on Baltimore City's DOT's website with this [new route here](https://transportation.baltimorecity.gov/charm-city-circulator/developer-resources). I have updated `static_current` to point to this new download URL, and added the previously used URL to `static_historic`. I don't see any other relevant changes in the feed, either

For reference: old feed's `routes.txt` on the left, new feed's `routes.txt` on the right
image

### DIFF
--- a/feeds/baltimorecity.gov.dmfr.json
+++ b/feeds/baltimorecity.gov.dmfr.json
@@ -5,8 +5,9 @@
       "id": "f-dqcx8-charmcitycirculator",
       "spec": "gtfs",
       "urls": {
-        "static_current": "https://transportation.baltimorecity.gov/gtfsfile",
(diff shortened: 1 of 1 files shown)


BNewborn:
I'm not exactly sure why this error is being thrown - is this something to do with Transitland's organization needing to update a billing setting on their side?

drewda:
Very occasionally GitHub will fail to launch jobs. After manually rerunning them, it looks fine now.

Thanks for this update.

PR #1459, opened 2025-02-28
Update solweb.tper.it.dmfr.json

Altonss:
Unfortunately they seem to update the link quite frequently ^^


drewda:
@Altonss thank you for this update

@jbruechert is correct that the issue is with the final line of the file. The "opinionated" DMFR format doesn't include a newline at the end of the file.

PR #1460, opened 2025-02-28
Update delijn.be.dmfr.json

sryyyl-tech:
There's an error while fetching actual DeLijn GTFS, so there's another link, working correctly and up-to-date. As a Belgian and working on a new public transport app, i can help you as needed: c.nomlus@gmail.com


sryyyl-tech:
Okay, found the error with the validation!

drewda:
@sryyyl-tech thanks for this contribution. Transitland also sources RT endpoints for De Lijn, so we ideally want to get both RT and static from the same host.

It looks like there may be some internal issues with Transitland using its existing authentication configuration to fetch this feed. We'll investigate further.

sryyyl-tech:
Hi!
Okay, I understand.
I'm making a Belgian app for public transport and it's a problem that DeLijn feed was not updated from 1 month.
Thanks for your work. Can you keep me informed?
If needed in the future, I can inform you if there's any changes with Belgian operators.

drewda:
Our team has fixed an internal issue with the fetch process, and https://www.transit.land/feeds/f-u1-delijn now shows successful fetches and imports again from the existing URL.

sryyyl-tech:
Thanks for your work! 🙂

PR #1461, opened 2025-03-03
Update realtime link for Valley Metro Phoenix

kylerchin:


kylerchin:
solved by echo "" >> feeds/transitdata.phoenix.gov.dmfr.json, since Ubuntu refuses to install transitland lib

drewda:
@kylerchin thanks for this update

kylerchin:
Over the next few weeks, the changes from https://github.com/catenarytransit/transitland-atlas will be moved one by one here.

Issue #1468, opened 2025-03-06
modify all DMFR files to have trailing line breaks


drewda:
Coordinate with the release of https://github.com/interline-io/transitland-lib/pull/416

This should make editing easier for people who are using the github.com interface to open PRs.

PR #1483, opened 2025-03-13
use transitland-lib v1.0.0, with change to trailing line break

drewda:
related to #1468


drewda:
closing and handling directly in #1482

PR #1491, opened 2025-03-19
Re-add Greenlink GTFS-RT feeds

wflake:
Greenlink GTFS-RT feeds were removed when updating the outdated static feeds. The new URLs are updated and current.


drewda:
@wflake thanks for this addition

PR #1508, opened 2025-04-11
Update gtfs.ovapi.nl.dmfr.json

jjasloot:
Added license and real time info to ovapi.nl

License also show on this page: https://openov.nl/

afbeelding

Small issue is that there exists both a tripUpdates.pb and a trainUpdates.pb file for this feed. Picked the trainUpdates for now. See http://gtfs.ovapi.nl/nl/

### DIFF
--- a/feeds/gtfs.ovapi.nl.dmfr.json
+++ b/feeds/gtfs.ovapi.nl.dmfr.json
@@ -6,6 +6,23 @@
       "spec": "gtfs",
       "urls": {
         "static_current": "http://gtfs.ovapi.nl/gtfs-nl.zip"
+      },
+      "license": {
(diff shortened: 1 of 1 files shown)


drewda:
@jjasloot thank you for these contributions.

In order to add an RT feed, it's also necessary to associate it with operator records.

Do you know which of the following agencies in the NL static feed have associated RT updates?

CleanShot 2025-04-14 at 11 24 24@2x

(This is a screenshot of https://www.transit.land/feeds/f-u-nl )

If so, then we should create operator records for those in the DMFR file and associate them with the new RT feed record.

jjasloot:
Are you refering to the `realtime_vehicle_positions` url or the `realtime_trip_updates` url?

PR #1513, opened 2025-04-14
Update license in data.pid.cz.dmfr.json

MartinNemi03:
Noticed that the license is wrong for this datasource as it should be "CC-BY-4.0" listed in "Licence" section of [PID's Opendata page](https://pid.cz/o-systemu/opendata/).

### DIFF
--- a/feeds/data.pid.cz.dmfr.json
+++ b/feeds/data.pid.cz.dmfr.json
@@ -8,9 +8,8 @@
         "static_current": "http://data.pid.cz/PID_GTFS.zip"
       },
       "license": {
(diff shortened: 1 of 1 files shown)


feeds/data.pid.cz.dmfr.json
```diff
@@ -27,5 +26,5 @@
       ]
     }
   ],
-  "license_spdx_identifier": "CDLA-Permissive-1.0"
```
drewda, line 30: `license_spdx_identifier` should not be changed. This refers to the DMFR file itself, rather than the individual feeds that are referenced.

PR #1595, opened 2025-05-19
added cityofanderson.com.dmfr.json (Electric City Transit)

blu3r4d0n:


drewda:
@blu3r4d0n thanks for this contribution

blu3r4d0n:
@drewda of course! I'm not sure if the month on that URL iterates every month but if it does we can revisit.

PR #1691, opened 2025-08-07
Update suffolkva.us.dmfr.json

szjozsef:
Suffolk Transit - url taken from: https://www.suffolkva.us/Search?searchPhrase=gtfs&pageNumber=1&perPage=10&departmentId=-1


drewda:
thank you for another update!

PR #1734, opened 2025-08-27
New Agencies in Great Lakes + South Central

BNewborn:
Reviewing our new agency requests for this year, I found 7 agencies with GTFS available online we could add to transitland. I've added them here following dmfr formatting and removed 6 from `2023-gtfs-weblinks.dmfr.json` (Murfreesboro was not in this json)


BNewborn:
I'll work on reformatting these

BNewborn:
Some notes while I do so

I'm not exactly sure of the best operator name to use for some of these - I left these as mirroring the NTD as much as possible. However, some of these operator names differ substantively from the transit agency name as we know it
| dba Agency Name | Operator Name | 
| ---- | ----- |
| Cass Area Transit / Cass Transit | Cass County Council on Aging | 
| Country Roads Transit | Randolph County Senior Center| 
| Ride Solution / WATS | Four Rivers Resource Services |
| City of Washington | City of Washington (however, this system is actually operated by Ride Solutions, though it feels distinct from WATS as each has their own NTD ID) | 

Open to any suggestions on how to rename these, if there are any concerns

drewda:
@BNewborn thanks for all these additions.

In 2 cases, the URLs were being "protected" by URL monitoring service, so I replaced those with the fully resolved URLs.

In terms of names, we have often had those questions about whether to use rider-facing brand names or formal agency names. In general, we lean to rider-facing brand names. When it's worth having both, we'll sometimes put the brand name in `short_name` and the agency name in `name`.

PR #1738, opened 2025-09-05
Transport Canberra update

applecuckoo:
self-explanatory, fixes the DMFR file for Transport Canberra to point towards their new platform as described [here](https://www.transport.act.gov.au/contact-us/information-for-developers).


drewda:
@applecuckoo thank you for this update

PR #1739, opened 2025-09-05
Update geodata.gov.gr.dmfr.json

alexroumi:
Add the new static current zip url

### DIFF
--- a/feeds/geodata.gov.gr.dmfr.json
+++ b/feeds/geodata.gov.gr.dmfr.json
@@ -8,7 +8,7 @@
       ],
       "spec": "gtfs",
       "urls": {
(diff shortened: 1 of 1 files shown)


Lach-anonym:
Greetings, the feed is expired

Lach-anonym:
No, it's about the calendar date, the last valid day is 18.02.2025
The separate files together from the same source give the GTFS with a later validity period, but in the end it's also expired
I haven't found a valid feed from Athens sadly

alexroumi:
Oh ok, in greece public transport is a bit lazy, i can assure you this is the valid one since it is the new platform they built for this kind of open source data, i will also be attatching a screenshot of the publish date of this zip. So i believe you can just ignore the validity date for now.
image

drewda:
Thanks for this update. It's better than what is currently in Atlas, so merging for now. And we'll hope for a new update soon.

PR #1746, opened 2025-09-15
Update brampton.ca.dmfr.json

useless2764:
Updated real time feeds to current urls


drewda:
Thank you for this update

PR #1782, opened 2025-11-03
Update Hamilton unstable URL value

tqpcharlie:
URL is no longer unstable.


drewda:
@tqpcharlie thanks for noticing

PR #1939, opened 2026-02-27
Add Brno GTFS RT URLs

MrPaschenko:


kudlav:
This GTFS RT does not contain `realtime_trip_updates`

PR #2017, opened 2026-05-06
Update for Triton Transit feed in-housing

42A7C5:
This change would move Triton Transit (UC San Diego Transportation Services shuttles) from the Transloc feed to the new in-house feed for both static and realtime.


42A7C5:
All checks should be passed now

drewda:
@42A7C5 thank you for this update!

PR #2022, opened 2026-05-07
Add realtime from M Rézo Grenoble.

kalon33:
@drewda Here I add an official GTFS-RT feed, listed in transport.data.gouv.fr but not in the proper place, so it's not captured by scripts parsing the website (see https://transport.data.gouv.fr/datasets/horaires-theoriques-du-reseau-tag).


kalon33:
They seem to have broken the feed. So for now, this commit is useless sorry...

PR #2058, opened 2026-06-19
Update static_current URL for transit feed

joshuahouston:
www.sitka.land has a firewall blocking mircrosoft azure (or something similar) so transit.sitka.land was setup to circumvent that as it's hosted on github


drewda:
@joshuahouston thank you for this update!

PR #2059, opened 2026-06-23
Update Puget Sound agencies to primary sources

sberkley:
* Update King County Metro to use their published data directly, rather than through OBA
* Update Everett Transit to use their published data directly, rather than through OBA
* Remove Seattle Children's Hospital feed listing since it is no longer included in OBA and isn't published publicly


drewda:
@sberkley thank you for these updates

PR #2066, opened 2026-07-10
Automated updates: URL updates from National Transit Database, websites and tags from Wikidata and scraping

impiaaa:
Initially I just wanted to update Visalia Transit, but that lead me to make a script that updates everything from NTD. But I figured I shouldn't add those without external links, so I also wrote a script to find websites and Wikidata items, and then use Wikidata and website scraping to get social links. But that script wouldn't work with duplicated or missing Onestop IDs, so I fixed those manually.

I manually checked over the diff to make sure that the websites all make sense, but I didn't validate that they all exist.


impiaaa:
I removed the two feeds that returned HTTP 404. `f-kalkaska~public~transit~authority` exists but is malformed (has a file named `agency (2).txt`), and I don't know how to fix that in this database. `f-paducah~transit~authority` and `f-texas~a~m~university` fail due to SSL cert validation, but succeed when ignoring the validation, and I don't know how to tell the validator to do that.

drewda:
Hi @impiaaa, thank you for this contribution and for sharing more about your process.

Please decompose this into multiple PRs that are more specific in what they change, to make it easier to review.

For instance, you could have one PR that address social media tags. Also when proposing changes to feed URLs, it's ideal if it's in a PR just focused on one feed or on a couple feeds that are closely associated by location, by domain name, or have something else in common.

PR #2087, opened 2026-07-31
Update Chattanooga GTFS Static

RJMcBug:
Updated CARTA GTFS feed as it moved from gocarta.org to GitHub


drewda:
Thanks for this update!

PR #1256, opened 2024-10-03
Onestop IDs shouldn't have capital letters

drewda:

PR #2099, opened 2026-08-10
Lowercase remaining uppercase operator Onestop IDs & fix some Greek feeds

drewda:
Onestop IDs are lowercase by convention, and #1256 made a pass at this in 2023, but twelve operator records survived it. All twelve carry uppercase in non-Latin scripts, which is presumably why a Latin-oriented sweep missed them: Greek, Cyrillic, Polish, Hungarian, Lithuanian, German and Irish, plus one Latin-script holdout in a hand-contributed Japanese file.

Each keeps its former id in `supersedes_ids`, so existing references still resolve.

## Checked before applying

Lowercasing non-Latin text is not automatically safe, so each was verified: **no length change, still NFC, idempotent**, and satisfying the Onestop format rules afterwards. Each id appeared exactly once in the registry, so there were no cross-references to update.

## Not caused by the generating scripts

All four generators that mint Onestop IDs already lowercase — `convert-gbfs-csv-to-dmfr.py`, `collect-gtfs-data-jp.py`, `collect-odpt-gtfs.py`, `collect-nap-gtfs.py`. These twelve are in hand-contributed files, so no generator change is needed.

## Two homoglyphs, fixed here rather than later

Both Greek operators mixed scripts mid-word, in the id and in the name. Since the ids were being rewritten anyway, fixing them now avoids superseding the same id twice:

- Athens wrote the English half of its name with a **Greek capital alpha**, `Αthens Urban Transport Organisation` (U+0391 rather than U+0041), which carried into the id as a Greek small alpha. Both are now Latin.
- Zakynthos is ΚΤΕΛ Ν. ΖΑΚΥΝΘΟΥ Α.Ε., but wrote ΖΑΚΥΝΘΟΥ with a **Latin capital Z** (U+005A rather than Greek zeta U+0396), which carried into the id as a Latin `z` between Greek nu and alpha. Both are now Greek zeta.

The names are the evidence for each reading and are corrected alongside the ids. The matching feed ids are brought into line too — an earlier pass had already lowercased feed ids but not operator ids, so leaving the feeds alone would have left the pair disagreeing.

Also corrects Patras, which spelled Προαστιακός ("suburban") as Προστιακός in its feed id, operator id and operator name; the word is προ-αστιακός, so the α is not optional. Adds `wikidata_id` Q3536559 to that operator. No `wikidata_id` for the others: every search for ΟΑΣΑ returns Q1814382, but that is instance-of "public transport network" and describes the Athens network rather than the organisation running it, so tagging the operator with it would assert something false. ΚΤΕΛ Ζακύνθου and ΚΤΕΛ Θεσπρωτίας have no entity at all.

## Greek URLs, forced by CI

Renaming these feed ids made the URL check validate their URLs for the first time — it keys its changed-set on feed id as well as URL, so a rename presents every URL on the feed as new. All four Greek feeds turned out to be dead, and none has ever produced a feed version.

Athens now points at the Ο.ΣΥ. export on `data.gov.gr`, the Greek national open data portal, published by ΟΑΣΑ itself: 485 routes, 97,217 trips, 8,040 stops, calendar 20260706 to 20261006. It validates with no errors. The dead `catalog.growthfund.gr` URL joins the older `geodata.gov.gr` one in `static_historic`.

Patras, Zakynthos and Thesprotias have no replacement — their 2017 datahub.io URLs 404 and nothing in the Mobility Database, the Greek portal or the operators' own sites offers another. All three operators still run, so the URLs move to `static_historic` rather than the records being deleted. Patras also gets its website updated, since trainose.gr now redirects to hellenictrain.gr.

Worth knowing for later: the portal's canonical resource URL for the Athens dataset ends `.rar` and really does serve a RAR archive, which transitland cannot read. Only the `.zip` variant is usable, and a made-up extension 404s, so the endpoint is not simply extension-agnostic.

## Why it matters

The registry now has **zero uppercase Onestop IDs across 2,728 operators and 6,221 feeds**. That means the lowercase check can be enforced on operators as well as feeds, rather than exempted — which is what interline-io/tlv2#489 asks transitland-lib to support, and what the Atlas-side validator currently has to skip for operators.

## Test plan

- `transitland dmfr format` and `dmfr lint` clean on all changed files.
- `validate-feeds.py` passes; `check-jsonschema` passes against the v0.6.0 schema.
- Re-scan of every DMFR file confirms zero uppercase ids remain, operators or feeds.
- `transitland validate` on the new Athens source reports no errors.

### DIFF
--- a/feeds/armadina.hu.dmfr.json
+++ b/feeds/armadina.hu.dmfr.json
@@ -15,7 +15,10 @@
       },
       "operators": [
         {
-          "onestop_id": "o-u2mmz-budaörsvárosÖnkormányzata",
+          "onestop_id": "o-u2mmz-budaörsvárosönkormányzata",
+          "supersedes_ids": [
+            "o-u2mmz-budaörsvárosÖnkormányzata"
+          ],
           "name": "Budaörs Város Önkormányzata",
           "website": "http://www.budaors.hu/"
         }
--- a/feeds/datahub.io.dmfr.json
+++ b/feeds/datahub.io.dmfr.json
@@ -12,7 +12,10 @@
       },
       "operators": [
         {
-          "onestop_id": "o-sgg5z-gr~ΠροστιακόςΣιδηρόδρομοςΠατρών",
+          "onestop_id": "o-sgg5z-gr~προστιακόςσιδηρόδρομοςπατρών",
+          "supersedes_ids": [
+            "o-sgg5z-gr~ΠροστιακόςΣιδηρόδρομοςΠατρών"
+          ],
           "name": "Προστιακός Σιδηρόδρομος Πατρών",
           "short_name": "Suburban Railway of Patras",
           "website": "https://www.trainose.gr/en/passenger-activity/suburban-railway/patras-suburban-railway/",
@@ -35,7 +38,10 @@
       },
       "operators": [
         {
-          "onestop_id": "o-sqz-greece~ΚΤΕΛΝzΑΚΥΝΘΟΥΑΕ",
+          "onestop_id": "o-sqz-greece~κτελνzακυνθουαε",
+          "supersedes_ids": [
+            "o-sqz-greece~ΚΤΕΛΝzΑΚΥΝΘΟΥΑΕ"
+          ],
           "name": "ΚΤΕΛ Ν. ZΑΚΥΝΘΟΥ Α.Ε.",
           "website": "http://ktel-zakynthos.gr/",
           "associated_feeds": [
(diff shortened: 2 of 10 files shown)


drewda:
## Also fixed: the two homoglyphs

Since the ids were being rewritten anyway, fixing these now avoids superseding the same id twice later.

Both records mixed scripts **mid-word**, in the Onestop ID *and* in the name — and the names turned out to be the evidence for each reading:

| | was | now |
|---|---|---|
| Athens id | `…οασα~αthensurbantransportorga` (Greek α) | `…οασα~athensurbantransportorga` |
| Athens name | `ΟΑΣΑ - Αthens Urban Transport Organisation` | `ΟΑΣΑ - Athens Urban Transport Organisation` |
| Zakynthos id | `…greece~κτελνzακυνθουαε` (Latin z) | `…greece~κτελνζακυνθουαε` |
| Zakynthos name | `ΚΤΕΛ Ν. ZΑΚΥΝΘΟΥ Α.Ε.` | `ΚΤΕΛ Ν. ΖΑΚΥΝΘΟΥ Α.Ε.` |

Athens wrote the English half of its name with a **Greek capital alpha** (U+0391 rather than U+0041), which carried into the id as a Greek small alpha before `thensurbantransportorga`. Now Latin.

Zakynthos is ΚΤΕΛ Ν. ΖΑΚΥΝΘΟΥ Α.Ε., the intercity bus operator for Zakynthos, but wrote ΖΑΚΥΝΘΟΥ with a **Latin capital Z** (U+005A rather than Greek zeta U+0396), which carried into the id as a Latin `z` between Greek nu and alpha. Now Greek zeta.

Names are corrected alongside the ids, since they are what appears on the website.

`supersedes_ids` is unchanged — it lists the originally published uppercase form. The intermediate lowercase-only form exists nowhere outside this branch, so there is nothing to supersede for it.

Verified: neither id changes script mid-word any more, and `dmfr lint`, `validate-feeds.py` and `check-jsonschema` all pass.

Issue #2145, opened 2026-09-14
SacRT doesn't show nay routes, need updating


bestskillzzz:
Hello both sacramento rt static gtfs and real-time gtfs don't work with routes, if it helps this is stic uro: https://apps.sacrt.com/gtfs/srtd/google_transit.zip. Also these are real time gtfs feeds: https://bustime.sacrt.com/gtfsrt/alerts, https://bustime.sacrt.com/gtfsrt/vehicles, and https://bustime.sacrt.com/gtfsrt/trips

PR #2152, opened 2026-09-15
Update sacrt.com Static GTFS schedule feed links

bestskillzzz:
Hello I updated the SacRT static GTFS link as it seems it was changed/updated from before


bestskillzzz:
Will close: https://github.com/transitland/transitland-atlas/issues/2145

